PR #1, opened 2019-04-09
Update README.md

JonnyBurger:
The automatic way adds a `types` field, not a `typescript` one.

So in the manual one it should probably also be `types`.


satya164:
Thank you!

PR #5, opened 2019-06-25
feat(aar): added configurable bundle name

Johan-dutoit:
# Summary

This adds the ability to change the aar bundle name bob uses.  Instead of always expecting `android.aar` (it's still the default).


satya164:
@matt-oakes wdyt?

PR #6, opened 2019-07-13
improve doc

slorber:
I guess there is no src/index.js because we are writing TS ;)

it's also found here: https://github.com/react-navigation/core/blob/master/package.json#L6

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ To configure your project manually, follow these steps:
        ["aar", {"reverseJetify": true}],
        ["commonjs", {"flow": true}],
        "module",
-       "typescript",
+       "typescript"
      ]
    }
    ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -62,7 +62,7 @@ To configure your project manually, follow these steps:
    ```json
    "main": "lib/commonjs/index.js",
    "module": "lib/module/index.js",
-   "react-native": "src/index.js",
+   "react-native": "lib/module/index.js",
```
satya164: Thanks for the PR. This should be `index.tsx` for TypeScript projects (unless you're supporting old versions of React Native without TypeScript support):

```suggestion
   "react-native": "src/index.tsx",
```

satya164:
Closing as the example is now fixed.

PR #19, opened 2020-01-28
fixed `types` in package.json generated by `bob init`

roothybrid7:
If you perform `bob init`,  it's written `lib/typescript/src/index.d.ts` to `types` in package.json. 
However, it's a different path that actually outputs TypeScript type definitions.

When you perform `bob build`, it will be output to `lib/typescript/index.d.ts`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ To configure your project manually, follow these steps:
    "main": "lib/commonjs/index.js",
    "module": "lib/module/index.js",
    "react-native": "src/index.js",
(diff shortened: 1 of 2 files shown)


satya164:
Hey! Thanks for the PR.

This path can actually change based on the project structure. I have added a template to generate a library which produces a path with `src` in it.

So it's up-to the maintainer to ensure that this path is correct. I have added a note mentioning it in the README.

In the future, I'd like add some lint task which can verify these things.

PR #21, opened 2020-02-08
feat: add abilty to generate a react-native project

satya164:
Kapture 2020-02-09 at 4 49 57


templates/library/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+```js
+import <%= project.name %> from "<%= project.slug %>";
+
+// ...
+
+const deviceName = await <%= project.name %>.getDeviceName();
+```
+
+## Scripts
```
thymikee: Put this into contributing guide

templates/library/_watchmanconfig
```diff
@@ -0,0 +1 @@
+{}
```
thymikee: I think this is not necessary anymore for watchman

PR #27, opened 2020-03-07
feat: added a non-native module library option

gorhom:
# Summary
Hi there 👋

I notice this library will always add the native module with integration in the example, which is great, but however, not all libraries will include the native module. 

So, i added the option when creating a new library to opt-in the native module with the example integration.

## Test Plan
Create new library > input `N` for `Does your library include native module/component?` > run example

image

Created Test Project ( removed native module integration )
image

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist

### DIFF
--- a/src/create.ts
+++ b/src/create.ts
@@ -1,6 +1,5 @@
 import path from 'path';
 import fs from 'fs-extra';
-import ejs from 'ejs';
 import dedent from 'dedent';
 import chalk from 'chalk';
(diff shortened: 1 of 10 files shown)


gorhom:
@satya164, yeah i think that's good idea, however, should we give an option to opt-in the example to be in expo or basic/raw ?

PR #30, opened 2020-03-23
feat: added possibility to start module with C++ code enabled

Patys:
# Summary
 - Script ask you do you want to add C++ code to your module (by default set to "No")
- It adds C++ files and additional code to load it to RN module
- you can use code inside cpp folder to develop your module

## Test Plan

1. Run script
2. Answer last question with Yes (Do you want to use C++ code?)
3. Example contains multiply code

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist

- [x] I added the documentation in `README.md`
- [x] I mentioned this change in `CHANGELOG.md`
- [x] I added a sample use of the API in the example project (`example/App.js`)


templates/cppLibrary/android/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
+cmake_minimum_required(VERSION 3.4.1)
+
+set (CMAKE_VERBOSE_MAKEFILE ON)
+set (CMAKE_CXX_STANDARD 11)
+
+add_library(cpp
```
gengjiawen, line 6: Maybe choose another name for library. cpp maybe a little confusing.

templates/cppLibrary/android/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
+cmake_minimum_required(VERSION 3.4.1)
+
+set (CMAKE_VERBOSE_MAKEFILE ON)
+set (CMAKE_CXX_STANDARD 11)
```
gengjiawen, line 4: Maybe 14 ? 11 is bit of old.

templates/cppLibrary/android/CMakeLists.txt
Patys: Actually it's a good idea. I will try to address that in next pr 👍

templates/cppLibrary/android/CMakeLists.txt
Patys: I had issues with newer versions as it's compiled for ios and android. I will try to investigate it more in future to bump it.

PR #31, opened 2020-03-24
feat: add option to specify tsc path

dpnolte:
# Summary
For the typescript target, this PR aims to add an option to specify a path to tsc. We are using bob in a workspace setup. With this option, we can resolve tsc's path to the workspace root's node modules so that not the global tsc is being used. 

## Test Plan

I've tested this fork with our setup and it works.

## Checklist

- [X ] I have tested this on a device and a simulator
- [X] I added the documentation in `README.md`
- [N/A ] I mentioned this change in `CHANGELOG.md`
- [ X] I updated the typed files (TS and Flow)
- [N/A ] I added a sample use of the API in the example project (`example/App.js`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,7 +170,7 @@ Example:
 
 Enable generating type definitions with `tsc` if your source code is written in [TypeScript](http://www.typescriptlang.org/).
 
-By default, it'll use the `tsconfig.json` file in your project root. If you want to use a different config, you can specify it using the `project` option. Furthermore, the tsc binary will be resolved to ./node_modules/.bin/tsc. Use the `tsc` option to sepcify a different path.
+By default, it'll use the `tsconfig.json` file in your project root. If you want to use a different config, you can specify it using the `project` option. Furthermore, the tsc binary will be resolved to ./node_modules/.bin/tsc. Use the `tsc` option to specify a different path.
 
 Example:
 
--- a/README.md
+++ b/README.md
@@ -170,7 +170,7 @@ Example:
 
 Enable generating type definitions with `tsc` if your source code is written in [TypeScript](http://www.typescriptlang.org/).
 
-By default, it'll use the `tsconfig.json` file in your project root. If you want to use a different config, you can specify it using the `project` option.
+By default, it'll use the `tsconfig.json` file in your project root. If you want to use a different config, you can specify it using the `project` option. Furthermore, the tsc binary will be resolved to ./node_modules/.bin/tsc. Use the `tsc` option to sepcify a different path.
 
 Example:
 
--- a/src/targets/typescript.ts
+++ b/src/targets/typescript.ts
@@ -78,10 +78,9 @@ export default async function build({
       );
     }
 
-    const tscPath = options?.tsc
-      ? options.tsc
-      : path.join(root, 'node_modules', '.bin', 'tsc');
-    let tsc = tscPath + (platform() === 'win32' ? '.cmd' : '');
+    let tsc = options?.tsc
+      ? path.resolve(root, options.tsc)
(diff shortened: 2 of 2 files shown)


src/targets/typescript.ts
```diff
@@ -78,9 +78,10 @@ export default async function build({
       );
     }
 
-    let tsc =
-      path.join(root, 'node_modules', '.bin', 'tsc') +
-      (platform() === 'win32' ? '.cmd' : '');
+    const tscPath = options?.tsc
+      ? options.tsc
+      : path.join(root, 'node_modules', '.bin', 'tsc');
+    let tsc = tscPath + (platform() === 'win32' ? '.cmd' : '');
```
satya164: I think if the user specified the path to `tsc`, we probably shouldn't modify it in any way to make it flexible (e.g. append `.cmd`). They can conditionally specify this path according to platform using `bob.config.js`.

src/targets/typescript.ts
```diff
@@ -78,9 +78,10 @@ export default async function build({
       );
     }
 
-    let tsc =
-      path.join(root, 'node_modules', '.bin', 'tsc') +
-      (platform() === 'win32' ? '.cmd' : '');
+    const tscPath = options?.tsc
+      ? options.tsc
```
satya164: If a relative path is specified, we should probably resolve it to be relative to `root`

src/targets/typescript.ts
dpnolte: appending `.cmd` is now only done when no tsc option is provided

src/targets/typescript.ts
dpnolte: I've added  a path resolve in my latest commit: 
```
let tsc = options?.tsc
      ? path.resolve(root, options.tsc)
      : path.join(root, 'node_module') + (platform() === 'win32' ? '.cmd' : '');
```

dpnolte:
Thanks for your feedback! I've incorporated your comments in the latest commit.
Also, as a fellow react-native developer, I'd like to say thank you for all the work you're putting into the react native community 🥇

dpnolte:
i've also updated the error/warning messages

README.md
```diff
@@ -170,7 +170,7 @@ Example:
 
 Enable generating type definitions with `tsc` if your source code is written in [TypeScript](http://www.typescriptlang.org/).
 
-By default, it'll use the `tsconfig.json` file in your project root. If you want to use a different config, you can specify it using the `project` option.
+By default, it'll use the `tsconfig.json` file in your project root. If you want to use a different config, you can specify it using the `project` option. Furthermore, the tsc binary will be resolved to ./node_modules/.bin/tsc. Use the `tsc` option to sepcify a different path.
```
jamesgeorge007, line 173:
```suggestion
By default, it'll use the `tsconfig.json` file in your project root. If you want to use a different config, you can specify it using the `project` option. Furthermore, the tsc binary will be resolved to ./node_modules/.bin/tsc. Use the `tsc` option to specify a different path.
```

src/targets/typescript.ts
```diff
@@ -78,9 +78,9 @@ export default async function build({
       );
     }
 
-    let tsc =
-      path.join(root, 'node_modules', '.bin', 'tsc') +
-      (platform() === 'win32' ? '.cmd' : '');
+    let tsc = options?.tsc
+      ? path.resolve(root, options.tsc)
+      : path.join(root, 'node_module') + (platform() === 'win32' ? '.cmd' : '');
```
jamesgeorge007, line 83:
```suggestion
      : path.join(root, 'node_modules') + (platform() === 'win32' ? '.cmd' : '');
```

PR #34, opened 2020-03-26
Update README: Fix types directory in the entry points example

DaniAkash:
# Summary
The example code in README leads to types not being identified after the package is published.
Causes issues like https://github.com/react-native-community/react-native-device-info/issues/814#issuecomment-535071998 & https://github.com/DaniAkash/react-native-responsive-dimensions/releases/tag/v3.1.1

## Test Plan
Not needed

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅   |
| Android |    ✅     |

## Checklist

- [x] I added the documentation in `README.md`

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,7 +94,7 @@ To configure your project manually, follow these steps:
    "main": "lib/commonjs/index.js",
    "module": "lib/module/index.js",
    "react-native": "src/index.ts",
-   "types": "lib/typescript/src/index.d.ts",
+   "types": "lib/typescript/index.d.ts",
    "files": [
      "lib/",
      "src/"


satya164:
Thanks for the PR. As mentioned in the `README.md`, this path will be different depending on the project configuration and layout.

> If you're building TypeScript definition files, also make sure that the types field points to a correct path. Depending on the project configuration, the path can be different for you than the example snippet.

Whatever we specify in the example won't work for 100% of the cases. So I'm closing this.

DaniAkash:
> Thanks for the PR. As mentioned in the `README.md`, this path will be different depending on the project configuration and layout.
> 
> > If you're building TypeScript definition files, also make sure that the types field points to a correct path. Depending on the project configuration, the path can be different for you than the example snippet.
> 
> Whatever we specify in the example won't work for 100% of the cases. So I'm closing this.

@satya164 This example will not work at all... that's the issue

satya164:
> @satya164 This example will not work at all... that's the issue

In your case with your project layout. Like I mentioned it's related to project layout. The same example works in my projects.

PR #37, opened 2020-04-03
Fix `types` path in template package.json

jamesisaac:
Building the `bob create` example places the transpiled contents of `src/` directly under the lib folders (`commonjs/`, `typescript/`, etc), i.e.:

`src/index.tsx` -> `lib/typescript/index.d.ts`

So the path to `types` in the default package.json, which includes `src/`, results in TS not being able to find types when the built library is consumed.

### DIFF
--- a/templates/library/$package.json
+++ b/templates/library/$package.json
@@ -4,7 +4,7 @@
   "description": "<%= project.description %>",
   "main": "lib/commonjs/index.js",
   "module": "lib/module/index.js",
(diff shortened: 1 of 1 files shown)


satya164:
Thanks for the PR, but this is not correct. The path is correct for the project generated with the template unless you have done any modifications.

jamesisaac:
Oh I see, it's because I deleted the `example` directory, so TS automatically expanded out the `src` directory I guess - sorry for the confusion.

PR #38, opened 2020-04-03
fix: add rn-config so CLI doesn't try to link it

jbtheard:
# Summary

Executing
```shell
pod install
```
currently results in pod complaining no pod spec exists for react-native-bob

This fix tentatively add a react-native-bob.podspec file that should solve this.

Note: this new file should be published at the root of the package. I'm not quite sure how this can be done. Please edit accordingly.

We also add a homepage property in package.json as it's required by standard podspec files

## Test Plan
When file manually added to node_module/@react-native-community/bob, executing pod install doesn't complain anymore

Not quite sure how to test before releasing a new version ?

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    n/a    |

## Checklist
- [X ] I have tested this on a device and a simulator


satya164:
Hmm, why would you need a podspec? This is not a native module. Seems like a bug in CLI?

thymikee:
Hey! It seems like CLI is detecting Bob as native library. It should be possible to disable autolinking of this package by creating a `react-native.config.js` in the root. Here's some documentation around the topic: https://github.com/react-native-community/cli/blob/master/docs/dependencies.md

jbtheard:
react-native.config.js file added as per @thymikee 's suggestion

PR #40, opened 2020-04-16
fix: exclude example and __tests__ in tsconfig

maxkomarychev:
# Summary

when running `yarn prepare` folder `lib/typescript` contains types
for all files in `example` as well as it includes `__tests__` from
library.

in order to avoid exampe and test types leak into disribution build is
made with another tsconfig which excludes mentioned files.

<details>
<summary>screenshots</summary>

before:
Screen Shot 2020-04-16 at 9 29 13 PM

after:
Screen Shot 2020-04-16 at 9 29 58 PM

</details>

## Test Plan

run `yarn prepare` with and without exclude.
ensure that `example` and `__tests__` are not included when exclude is present.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅❌     |
| Android |    ✅❌     |

## Checklist

### DIFF
--- a/src/targets/typescript.ts
+++ b/src/targets/typescript.ts
@@ -25,7 +25,7 @@ export default async function build({
 
   report.info(`Generating type definitions with ${chalk.blue('tsc')}`);
 
-  const project = options?.project ? options.project : 'tsconfig.json';
+  const project = options?.project ? options.project : 'build.tsconfig.json';
   const tsconfig = path.join(root, project);
 
   try {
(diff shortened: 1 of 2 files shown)


maxkomarychev:
That's sounds correct, how about we introduce another tsconfig and build using it? Could you please review new solution?

satya164:
I think it will be better to just delete these files after build since it'll also work for people who already had a generated project and people who are already using it without any extra setup.

So we can delete files matching this `glob` - https://github.com/react-native-community/bob/blob/7ee5ea2bfb85b617fe4ddefe2fc097ebf83d6145/src/utils/compile.ts#L29

And then delete all the extra folders except the `src` inside `lib/typescript`.

We could add an `exclude` option as well to configure this for all the targets (commonjs, module and typescript).

PR #48, opened 2020-05-08
feat: run linter only on modified files prior to making a commit

jamesgeorge007:
# Summary
Added [lint-staged](https://github.com/okonet/lint-staged) as a `devDependency` to ensure only those files which are modified is being taken into consideration by `eslint`. Hence, adding to the DX.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅❌     |
| Android |    ✅❌     |

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,6 +72,7 @@
     "eslint": "^6.8.0",
     "eslint-config-satya164": "^3.1.6",
     "husky": "^4.2.5",
+    "lint-staged": "^10.2.2",
     "prettier": "^2.0.5",
     "release-it": "^13.5.8",
     "typescript": "^3.8.3"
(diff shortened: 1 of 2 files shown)


satya164:
Thanks for the PR. IMO running eslint on all files is better since even changed files could also affect unchanged files. For example, for stuff like eslint-plugin-import, renaming/deleting/changing exports in a different file would cause errors in the file that's importing it. Updating dependencies such as ESLint, Prettier or the config package may also cause errors due to changes in rules even if no JS/TS files are changed, and there are others I may have missed.

For very large projects, it makes sense to use lint-staged since otherwise linting can be slow. But I think it doesn't make any difference for small projects like this one.

jamesgeorge007:
I went ahead and submitted the PR as it's a common practice to use husky and lint-staged together in a project. And as you said things would be different for small projects. In that case feel free to close this one.

jamesgeorge007:
Also, it seems e2e tests are missing out, I hope you're looking forward have them onboarded. Would love to collaborate 🙌

satya164:
Would be awesome if you want to work on it. I just added ability to generate projects by passing CLI args so it should be easy to generate projects on the CI.

PR #51, opened 2020-05-25
feat: add JS module option

gorhom:
# Summary

Hi there 👋

This is an updated closed pr (https://github.com/react-native-community/bob/pull/27) using templates style.

## Test Plan

Preview

Create new library > ... > select `JavaScript module` type > 🎉

nothing

nothing

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist


gorhom:
hi @satya164 , would you mind review this pr :)

satya164:
Thanks for the PR. Couple of things:

1. We need to reuse existing native code instead of duplicating it, since it increases work for updating, all of the templates reuse common code when they can
2. If it's JS only project, we also need web support (we should add it to the native one too)

Maybe it's easier just to init the expo project and then eject it if you need access to native code instead of separate template? It will already have web support that way.

gorhom:
hi @satya164 , thanks for getting back to my pr,

you are right, i think i should reuse the native code instead duplicating a whole template.

> If it's JS only project, we also need web support (we should add it to the native one too)

my only concern that not all non-native libraries could run on web, because whether its functionalities not suitable for web usage or its dependencies not compatible with `react-native-web` or `expo-web`.

if this still make since, please let me know and i'll fix this pr :)

thanks

satya164:
Yeah I agree, but it certainly depends on a case by case basis. If we include a webpack config by default, people can remove if they don't be it.

Anyway, a webpack config can be added in a follow-up PR, so it's fix the duplication in this PR and we'll merge.

src/create.ts
```diff
@@ -262,7 +267,15 @@ export default async function create(argv: yargs.Arguments<any>) {
...
 
-    if (type === 'cpp') {
+    if (type === 'js') {
+      await Promise.all(
+        [
+          `${folder}/android`,
+          `${folder}/ios`,
+          `${folder}/${options.project.podspec}.podspec`,
+        ].map(async (file) => await fs.remove(path.resolve(file)))
+      );
```
satya164, line 277: Few things:
1. Use `path.join` instead of string interpolation for cross-platform compat
2. Extra async/await in the map is not needed
3. Extra `path.resolve` is not needed since it's already full path

```suggestion
      await Promise.all(
        [
          `android`,
          `ios`,
          `${options.project.podspec}.podspec`,
        ].map((file) => fs.remove(path.join(folder, file)))
      );
```

satya164:
Thanks!

PR #55, opened 2020-06-05
feat: add native swift option

SimonErm:
# Summary
This PR adds the option to create a native module with swift for iOS.
Adresses https://github.com/react-native-community/bob/issues/54

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

## Checklist

### DIFF
--- a/src/create.ts
+++ b/src/create.ts
@@ -65,7 +65,7 @@ export const args: Record<ArgName, yargs.Options> = {
   },
   'type': {
     description: 'Type package do you want to develop',
-    choices: ['native', 'swift-native', 'cpp', 'expo'],
+    choices: ['native', 'native-swift', 'cpp', 'expo'],
   },
 };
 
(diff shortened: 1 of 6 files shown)


src/create.ts
```diff
@@ -64,7 +65,7 @@ export const args: Record<ArgName, yargs.Options> = {
   },
   'type': {
     description: 'Type package do you want to develop',
-    choices: ['native', 'cpp', 'expo'],
+    choices: ['native', 'swift-native', 'cpp', 'expo'],
```
satya164, line 68: Looks like a typo? Above it's `native-swift`

SimonErm: Yes you're right. I just tested the project init with the wizard, so i didn't recognize this.

src/create.ts
```diff
@@ -218,8 +220,9 @@ export default async function create(argv: yargs.Arguments<any>) {
         .slice(1)}`,
       package: slug.replace(/[^a-z0-9]/g, '').toLowerCase(),
       podspec: slug.replace(/[^a-z0-9]+/g, '-').replace(/^-/, ''),
-      native: type === 'native' || type === 'cpp',
+      native: type === 'native' || type === 'cpp' || 'swift-native',
```
satya164, line 223:
```suggestion
      native: type === 'native' || type === 'cpp' || 'native-swift',
```

PR #62, opened 2020-06-30
fix: Example Android app directory structure

Krizzu:
# Summary

Example android app created for `native-library` had wrong directory structure (Manifest was pointing wrong `package`). Changed directories to `com.example.%package-name%`.

Also, added `.idea` to `.gitignore`, in case someone using WebStorm (or similar) would try to contribute :)

## Test Plan

Create new `native library` and check paths for `example` android app. 
Try opening `example/android/app/build.gradle` with Android studio (no issues should be notified).

## Checklist

- [X] I have tested this on a device and a simulator


satya164:
Thanks!

PR #64, opened 2020-07-02
feat: native view managers

Krizzu:
# Summary

Updated templates to include examples for UI Managers in Kotlin, ObjC and Swift. View displays a box, which can change its background color by passing `color` prop (hex color).

# Test plan

1. Build the latest `bob`, generate project with `Kotlin/Obj` or `Kotlin/Swift` modules
2. Run example App, confirm the view is displayed.

## Screenshots

```react

<MyViewManager color="#32a852" style={styles.box} />

```

### Android
android

### iOS
ios

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅    |

## Checklist

- [X] I have tested this on a device and a simulator

### DIFF
--- a/templates/swift-library/ios/{%= project.name %}ViewManager.m
+++ b/templates/swift-library/ios/{%= project.name %}ViewManager.m
@@ -1,0 +1,7 @@
+#import "React/RCTViewManager.h"
+
+@interface RCT_EXTERN_MODULE(<%= project.name %>ViewManager, RCTViewManager)
+
+RCT_EXPORT_VIEW_PROPERTY(color, NSString)
+
+@end
--- a/templates/common/example/src/App.tsx
+++ b/templates/common/example/src/App.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import { StyleSheet, View, Text } from 'react-native';
-import <%= project.name %> from '<%= project.slug %>';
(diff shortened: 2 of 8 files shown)


templates/swift-library/ios/{%= project.name %}ViewManager.m
```diff
@@ -0,0 +1,7 @@
+#import "React/RCTViewManager.h"
```
yvbeek, line 1: Shoudn't this be `#import <React/RCTViewManager.h>`?

Krizzu: I think in this case the Header search paths are the same, so both "global" and "local" import work here

thymikee:
Great work! What do you think about adding ViewManagers as a separate command? I'm not sure how hard would it be to make it optional, @satya164 will know. But it's pretty rare for a library to include both native module and a view manager. With this change as is, users would get a lot of noise when they don't exactly need it. 

Ideally, there would be a possibility to add a view manager to the existing project bootstrapped with Bob as well, not only at the creation time.

Krizzu:
@thymikee Thanks!

IMO if we'd plan to make `ViewManagers` optional, we should do the same for `NativeModule` to not favour one over another. This could be done by adding another inquiry after selecting module type.

satya164 (approved):

PR #73, opened 2020-07-23
chore: update is-git-dirty

JPeer264:
# Summary

--

--

## Checklist


Krizzu:
Any reason for updating major version?

JPeer264:
The only breaking change is that I (as I am the author of this package) dropped Node versions below 10 and updated the dependencies. As this package only satisfies versions above v10 I updated `is-git-dirty` to v2.

PR #76, opened 2020-08-06
fix: fix Obj-C file extension in project.pbxproj

Simek:
# Summary

Currently when creating Native Module package for iOS the invalid Objective-C implementation file extension (`.mm`) is provided in `project.pbxproj`. This causes the inability to open and edit the package [`.m` file](https://github.com/react-native-community/bob/tree/master/templates/objc-library/ios) in XCode.

This PR fixes that issue simply by replacing the incorrect file extension with the correct one.

## Test Plan

After generating the project with the change I can open implementation code in the XCode.

XCode!

Generate iOS Native Project and try to open main file in XCode.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

## Checklist

### DIFF
--- a/templates/native-library/ios/{%= project.name %}.xcodeproj/project.pbxproj
+++ b/templates/native-library/ios/{%= project.name %}.xcodeproj/project.pbxproj
@@ -13,7 +13,7 @@
 <% if (project.swift) { %>
 		F4FF95D7245B92E800C19C63 /*  <%= project.name %>.swift in Sources */ = {isa = PBXBuildFile; fileRef = F4FF95D6245B92E800C19C63 /*  <%= project.name %>.swift */; };
 <% } %>
-		5E555C0D2413F4C50049A1A2 /* <%= project.name %>.mm in Sources */ = {isa = PBXBuildFile; fileRef = B3E7B5891CC2AC0600A0062D /* <%= project.name %>.mm */; };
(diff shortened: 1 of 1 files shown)


satya164 (changes requested):
Hey! Thanks a lot for the PR and sorry for the late review. We need `mm` extension for C++ modules afaik. Can you add a conditional to have `mm` for C++ module and `m` for rest like we do for swift?

Simek:
> Hey! Thanks a lot for the PR and sorry for the late review. We need `mm` extension for C++ modules afaik. Can you add a conditional to have `mm` for C++ module and `m` for rest like we do for swift?

@satya164 Thank you for the review! I have a stashed changeset with review changes ready, but I want to ask one question before I push it: Dose the Swift project requires `.m` files too? It is not obvious from the current file, because `.m` files are present in some swift conditionals but not all of them (PBXBuildFile section do not include source `.m` file for example).

satya164:
I had received a PR for adding Swift and C++ so I am not 100% sure, but if I understand correctly, for both Swift and Obj-C, we need `.m` and for C++ we need `.mm` :)

Simek:
I hope that I made the correct changes. I have tried to clean up the logic a bit by merging conditionals (so now there are always three branches: CPP, Swift and rest). Let me know what do you think and if updates are correct in the first place. 🙂

satya164 (approved):

PR #96, opened 2020-10-14
fix: eslint config in package.json

### LINKED ISSUE #86: Change the default ESLint config for common template to include "root": true
# Feature request
If this flag is not set, ESLint will automatically look for eslint config files in the directory of the file to be linted, and in successive parent directories all the way up to the root directory of the filesystem ([reference](https://eslint.org/docs/user-guide/configuring#using-configuration-files-1)). So if using your new package e.g. as a git submodule, the package's config can conflict with the parent project's config, causing ESLint to fail.

## Why it is needed
The above mentioned issue happened when running `yarn lint` for my new package, that was included as a git submodule for my main project. I can't think of any primary use cases where you would want the eslint config of this package to search for configs in parent directories, at least it shouldn't be the default behaviour IMO. 

## Possible implementation
Add the `"root": true` flag to the default ESLint config in the `package.json` for the common template.

### Code sample
https://github.com/anfriis/bob/commit/3a91f5f9abf200159edd6ff45b195b1c9340b8fb

sAleksovski:
# Summary

Fixes #86

## Test Plan

1. Build the latest bob, generate project with "JavaScript module with Expo example and Web support"
2. Verify `"root": true,` is present in generated `package.json file

## Checklist

### DIFF
--- a/templates/common/$package.json
+++ b/templates/common/$package.json
@@ -102,6 +102,7 @@
     }
   },
   "eslintConfig": {
+    "root": true,
     "extends": [
       "@react-native-community",
       "prettier"


Krizzu (approved):

Krizzu:
Thanks!

PR #99, opened 2020-10-25
Feature/configuration as files

SimonErm:
### Summary
This PRs adds the option to store the configurations of prettier, eslint, ... as seperate files instead of pacakge.json.


satya164:
Thanks for the PR. Personally I dislike having several configuration files in the root of the project if it can be avoided. However, ultimately it's stylistic choice. For Bob, we have decided to keep it opinionated and this means that we don't want to add options which are only a matter of opinion and don't affect the functionality.

So I'll be closing this PR.

PR #105, opened 2020-11-13
refactor: add separate option to generate view modules

### LINKED ISSUE #93: Additional command to select Native Module type
# Feature request

After merging [this feature](https://github.com/react-native-community/bob/pull/64), `bob` includes both Functional and UI native modules upon bootstraping. Not all projects require both types though, so additional command to select all/one/none would be great to have.

troZee:
### Summary

After merging [this feature](https://github.com/react-native-community/bob/pull/64), `bob` includes both Functional and UI native modules upon bootstraping. Not all projects require both types though, so additional command to select all/one/none would be great to have.

- Added a separate template folders (`native-view-library`, `objc-view-library`,  `swift-view-library`) to add only necessary files to the generated project 
- Added new type `ModuleType` to distinguish, what type of module dev would like to generate 
- The mixed case (view module with native module) has not been added here, because in most cases ppl create only a separate modules (We can consider to implement it in separate PR)

Closes #93 

### Test plan
- run below command for all types 

```sh
../bob/bin/bob create test-project
```

- run `yarn bootstrap` and then `yarn android` & `yarn ios`
- check, if all projects has been generated correctly 

### Screenshots 

Native View Module 
Screen Shot 2020-11-13 at 1 05 29 PM

Native module
Screen Shot 2020-11-13 at 12 51 01 PM

Added below two options:
Screen Shot 2020-11-13 at 3 00 33 PM


satya164 (approved):
Thank you!

Issue #101, opened 2020-10-28
ios project.pbxproj refers to .mm files rather than .m files


lastarrats:
Creating a package choosing Kotlin and Objective-C, project.pbxproj refers to .mm files instead .m files

Run:
npx @react-native-community/bob create react-native-awesome-module

? What type of package do you want to develop? 
❯ Native module in Kotlin and Objective-C 

cd ios/AwesomeModule.xcodeproj 
vi project.pbxproj 

-- All references refer to AwesomeModule.mm rather than AwesomeModule.m

cainaleaouk:
@lastarrats Doing a find and replace on `project.pbxproj` to fix the file extensions worked for me 👍

PR #111, opened 2020-12-10
Fix issue #101 pbxproj referrs to a .mm file instead of .m file

kochavi-daniel:
### Summary

This PR solves issue #101 where the project.pbxproj makes a reference to a `.mm` file instead of the `.m` file in the template when selecting a `Native module in Kotlin and Objective-C` project template.

### Test plan

Create a new project, open the project in XCode and the `.m` file should be available now from the project navigator.


satya164:
Thanks for the PR. However, this is already fixed in https://github.com/callstack/react-native-builder-bob/pull/76

PR #134, opened 2021-02-07
Fix EJS formatting issues

friederbluemle:
### Summary

A couple of related fixes to the EJS statements:

#### Remove leading spaces

Full length EJS lines should _not_ be indented as if they were semantically on same level as the code they encapsulate. Instead, they should be considered independent statements that _assemble indented code_. Otherwise the processed files will contain whitespace errors (i.e. spaces followed by EOL).

#### Slurp newlines

Change the end marker `%>` to `-%>` for full line ejs statements. What this does is it will _omit_ the newline character in the processed file. Depending on the conditional path, `yarn lint` (ESLint) was failing with a double blank line, e.g.:

```
[...]/example/src/App.tsx
  6:1  error  Delete `⏎`  prettier/prettier
```

### Test plan

Create new project and check generated files are whitespace error free. Verify `yarn lint` passes.

### DIFF
--- a/templates/common/example/src/App.tsx
+++ b/templates/common/example/src/App.tsx
@@ -1,12 +1,14 @@
 import * as React from 'react';
-<% if (project.moduleType === "view") { %>
+
+<% if (project.moduleType === "view") { -%>
 import { StyleSheet, View } from 'react-native';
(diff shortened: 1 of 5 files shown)


satya164:
Thanks!

Issue #145, opened 2021-03-09
Incorrect bob version in package.json file


troZee:
## Environment

MacOS

## Description

* `npx create-react-native-library react-native-awesome-module`
* `cd  react-native-awesome-module`
* `yarn bootstrap`

## Reproducible Demo

![Screen Shot 2021-03-09 at 11 36 29 AM](https://user-images.githubusercontent.com/12766071/110458554-2e18e800-80cc-11eb-851e-128d5a14058a.png)

Krizzu:
Hello @troZee 

Looks like [npm view](https://github.com/callstack/react-native-builder-bob/blob/main/packages/create-react-native-library/src/index.ts#L329-L336) returned empty response instead a version. Are you able to reproduce it every time?
I just tried couple of times and I get the correct version.

troZee:
Hello @Krizzu 
Thank you for checking this issue. I am still able to reproduce it on my end. 

Below I am providing more info to reproduce it
![Screen Shot 2021-03-11 at 12 48 30 PM](https://user-images.githubusercontent.com/12766071/110783394-addbb980-8268-11eb-936c-0619f13decef.png)

prolok:
I have seen this too and I have to manually add the react-native-builder-bob by calling

```
yarn add --dev react-native-builder-bob
```

PR #147, opened 2021-03-11
fix: fallback bob version

troZee:
### Summary

Closes https://github.com/callstack/react-native-builder-bob/issues/145

The fallback version was incorrect. This PR fixes it.

### Test plan

Like always :D

### DIFF
--- a/packages/create-react-native-library/src/index.ts
+++ b/packages/create-react-native-library/src/index.ts
@@ -343,7 +343,6 @@ async function create(argv: yargs.Arguments<any>) {
     ]);
   } catch (e) {
     // Fallback to a known version if we couldn't fetch
-    version = FALLBACK_BOB_VERSION;
   }
 
   const options = {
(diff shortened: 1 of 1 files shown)


packages/create-react-native-library/src/index.ts
```diff
@@ -341,7 +343,7 @@ async function create(argv: yargs.Arguments<any>) {
     ]);
   } catch (e) {
     // Fallback to a known version if we couldn't fetch
-    version = '0.18.0';
+    version = FALLBACK_BOB_VERSION;
```
Krizzu, line 346: You can leave this catch empty

Krizzu (approved):
Nicely 👍  This will make sure that we always have a fallback version, even if fetching version from npm takes longer than one sec

packages/create-react-native-library/src/index.ts
```diff
@@ -319,7 +321,7 @@ async function create(argv: yargs.Arguments<any>) {
       : 'module';
 
   // Get latest version of Bob from NPM
-  let version: string;
+  let version: string = FALLBACK_BOB_VERSION;
```
satya164, line 324: Instead of this, let's do `version: version || FALLBACK_BOB_VERSION` below where it's used. it'll make sure it works even if `version` is empty for some reason

PR #161, opened 2021-04-22
Fix wrong package name of ReactNativeFlipper

dsonet:
### Summary
When the Flipper is enabled, the generated code package name is wrong.
It will throw a runtime error since can't find the ReactNativeFlipper class.


Krizzu (approved):
Thanks

PR #164, opened 2021-05-01
feat: upgrade husky to v6

crutchcorn:
### Summary

The previous generated templates were using Husky v4. However, Husky v6 is the current major version. v5 introduces major breaking changes that [now require your FS to be structured in a particular way](https://blog.typicode.com/husky-git-hooks-javascript-config/). This PR updates the dependency for future React Native libraries

### Test plan

- Generate new project
- Confirm husky still functions as-intended


satya164 (approved):
Thank you for the PR. Still not a fan of this change cluttering project root 😅

jahead:
agreed, and thanks for the PR review

PR #165, opened 2021-05-05
fix: can't run yarn command under Windows

dsonet:
### Summary
can't run yarn command under Windows

### Test plan
At the root of the project, run yarn, it fails silently. (log the result will see the error is ENOENT.)
Add this option shell = true, it runs normally.


satya164 (approved):
Thank you!

crutchcorn:
This was driving me WILD and couldn't figure out why this was only happening in some projects and not others - thank you SO much for this

DiegoBM:
Thanks man, I was going mad. I hope they fix this soon

PR #168, opened 2021-05-21
`blacklist` -> `exclusionList` migration for examples metro.config.js

### LINKED ISSUE #160: blacklist is now exclusionList in metro 0.64+
## Environment

N/A

## Description

metro moved the blacklist file to exclusion list [here](https://github.com/facebook/metro/commit/3e2d9116d8a7f2580dcda51990cca5b3d98a81ff#diff-dc4b689d9c35d1bcb231f4251f49fe9e30d6656dc30dcd951e4d3289ea20d258).

The [metro.config.js template file](https://github.com/callstack/react-native-builder-bob/blob/main/packages/create-react-native-library/templates/common/example/metro.config.js) just needs an update to use 'metro-config/src/defaults/exclusionList' instead.

## Reproducible Demo

Any using metro 0.64+

BartoszKlonowski:
This pull request fixes #160 

It changes the *blacklist* to *exclusionList* in the example *metro.config.js* file so it is compatible with React Native v0.64.x.
It also updates the React Native version of example to v0.64.0

---

Two main changes are introduced in this delivery:
* Update the version of React Native to 0.64.0
This is done to ensure a compatibility with *exclusionList*.
Please note that once React Native Windows is supported it will most likely require the 0.64.x version of both RN and RNW.
* Change all occurrences of `blacklist` to `exclusionList` in *templates/common/example/metro.config.js*

### DIFF
--- a/packages/create-react-native-library/templates/common/example/metro.config.js
+++ b/packages/create-react-native-library/templates/common/example/metro.config.js
@@ -1,5 +1,5 @@
 const path = require('path');
-const blacklist = require('metro-config/src/defaults/blacklist');
+const exclusionList = require('metro-config/src/defaults/exclusionList');
 const escape = require('escape-string-regexp');
 const pak = require('../package.json');
 
@@ -14,9 +14,9 @@ module.exports = {
   watchFolders: [root],
 
   // We need to make sure that only one version is loaded for peerDependencies
-  // So we blacklist them at the root, and alias them to the versions in example's node_modules
+  // So we block them at the root, and alias them to the versions in example's node_modules
   resolver: {
-    blacklistRE: blacklist(
+    exclusionListRE: exclusionList(
       modules.map(
         (m) =>
           new RegExp(`^${escape(path.join(root, 'node_modules', m))}\\/.*$`)
--- a/packages/create-react-native-library/templates/common/example/metro.config.js
+++ b/packages/create-react-native-library/templates/common/example/metro.config.js
@@ -16,7 +16,7 @@ module.exports = {
   // We need to make sure that only one version is loaded for peerDependencies
   // So we block them at the root, and alias them to the versions in example's node_modules
   resolver: {
-    exclusionListRE: exclusionList(
+    blacklistRE: exclusionList(
       modules.map(
         (m) =>
           new RegExp(`^${escape(path.join(root, 'node_modules', m))}\\/.*$`)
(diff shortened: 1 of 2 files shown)


packages/create-react-native-library/templates/common/example/metro.config.js
```diff
@@ -14,9 +14,9 @@ module.exports = {
   watchFolders: [root],
 
   // We need to make sure that only one version is loaded for peerDependencies
-  // So we blacklist them at the root, and alias them to the versions in example's node_modules
+  // So we block them at the root, and alias them to the versions in example's node_modules
   resolver: {
-    blacklistRE: blacklist(
+    exclusionListRE: exclusionList(
```
walterholohan, line 19: I just tested this and this needs to be 
```
blacklistRE: exclusionList(
```

otherwise it doesn't work

BartoszKlonowski: Are you sure it's `blacklist` and not a `blockList`?
Please check 
[this line](https://github.com/facebook/metro/blob/1b905b60e397f2213f4ad20eb8201516b55affab/packages/metro-config/src/defaults/index.js#L39) in metro.

I can see that this is quite messy, as the history of the *packages/metro-config/src/defaults/index.js* contains two changes, and the issue which this pr closes points to the first one, which leaves the `blacklistRE` where it was, but another change 
[Add support for blockList and array of regexs ](https://github.com/facebook/metro/commit/94c0b541b4bfa17aee4efa0f1969565522ce830d#diff-1a3c1a959bb8c4e2e9743c03cb7a6d0c56648ffcfe129a11b9090bfc139622dd) makes it a `blockList` instead of `blacklist`.
Are you sure you have tested it on the latest of metro?

walterholohan: Yes im sure. Checkout my repo - https://github.com/walterholohan/react-native-crisp-chat-sdk/blob/master/example/metro.config.js.

Ive tested the above and it works

BartoszKlonowski: Sure! Then I will deliver the fix within a minute.
Thank you very much for your interest, @walterholohan!

myckhel (approved):
This is a great fix.

vonoro (approved):

PR #172, opened 2021-05-26
Support React Native Windows - C++/WinRT and C# templates with Example

### LINKED ISSUE #69: Add template for React Native Windows plugins
# Feature request

It would be nice to have a plugin template for [React Native Windows](https://github.com/microsoft/react-native-windows). Perhaps someone from Microsoft (@stmoy?) might be able to help out?

## Why it is needed

Adding a template for RN Windows will help potential contributors follow a "best-practice" way of adding support for Windows to their desired plugins

## Possible implementation

Windows projects use Visual Studio solution files (`*.sln`). The following repositories have Windows support:

- [`react-native-webview`](https://github.com/react-native-community/react-native-webview)
- [`async-storage`](https://github.com/react-native-community/async-storage)
- [`react-native-netinfo`](https://github.com/react-native-community/react-native-netinfo)

BartoszKlonowski:
This pull request closes #69 

It delivers the full support of React Native Windows project's templates for both C# and C++ languages.
Example project is configured as C# application.

---

Following changes were implemented:
* Additional questions and options in the *index.ts* file for Windows support
* *package.json* extended with *react-native-windows: v0.63.4*
* C++/WinRT project templates for both native module and native UI component
* C# project templates for both native module and native UI component
* README updated with the information about Windows support
* Example application provided with the Windows solution (*.sln*)

The 4 new project templates created in the Bob project are created as a 0.63.x projects with React Native Windows dependency already added to the solution configuration.
None of the scaffolded projects use the precompiled headers. All headers required for implementation are included where they are required.

**NOTE:** Each language has its project's GUID being the same across native module and native UI component.

All projects are provided with the most basic source code file, which should be a starting point for a further implementation.
Though they are basic and do nothing, they still compile successfully for each project.
Those files and classes they implement are named according to the type of project selected.
These basic modules/component are already registered in the *ReactPackageProvider*.

---

Please see the short presentation of the full template creation looks like (C# is taken as an example, but the procedure is of course the same for C++/WinRT):

BobExample

---

For some more explanations please check the commit messages.


asklar:
Love seeing the support for Windows apps being added. Main piece of feedback is that the project files and the code that the template RNW CLI produce, change over time. That is to say, an app created for 0.63 RNW will have different project/code files as 0.64 RNW. We expect we will continue to iterate on the template code and project files as we continue to evolve the platform. Sometimes we also update the template files/projects to backport a severe bugfix.

So ideally react-native-builder-bob wouldn't carry these files itself and instead just call the RNW CLI (or somehow "suck up" the files that RNW CLI uses for each version).

asklar:
@jonthysell to review too

BartoszKlonowski:
@asklar You've got the point - creating the project through the CLI would prevent us from updating the template and track backports. On the other hand current solution allows to do some customizations to the scaffolded solution, also whole system (each language, platform) is done like this so there's no confusion for a future maintainer, I also believe that the amount of changes required to be done to implement CLI for Windows only would be bigger than the current solution.
Of course, the other side of a coin is that the XML file can be more confusing than command line, but I'm aware of that and to be honest I'm willing to keep on maintaining that.

**But** I see no problem in keeping your opinion in mind, so perhaps whole system will be based on CLI, not just Windows.

asklar:
Yes I can see your point; I wouldn't want Windows to be the odd one out that's different : ) I just worry about how RNBB will be able to keep up/know about changes to the templates, but if you guys are comfortable signing up to keep this maintained then I am not concerned :)

BartoszKlonowski:
@asklar What just came to my mind is that there's one main concern about CLI-based approach:
*react-native-windows-init* have the `--projectType` flag which determines whether *exe* or *dll* will be created, but if it's the native module or native UI component is determined only by the files and the implementation of the project. So there would be no possibility to create those two separately.
I'm glad that we agree about the current approach, but CLI-based approach should be something to think about once it's possible for each platfrom and environment.

BartoszKlonowski:
@asklar @jonthysell Did you take a chance to take a look at .vcxproj and .csproj files? These are essential for the Builder Bob tool to work correctly and your expertise is most valuable here.

packages/create-react-native-library/src/index.ts
```diff
@@ -356,6 +388,7 @@ async function create(argv: yargs.Arguments<any>) {
       cpp: languages === 'cpp',
       kotlin: languages === 'kotlin-objc' || languages === 'kotlin-swift',
       swift: languages === 'java-swift' || languages === 'kotlin-swift',
+      windows: languages === 'windows-cpp' || languages === 'windows-csharp',
```
asklar, line 391: suggest that we call these uwp-cpp and uwp-csharp, or windows-uwp-cpp / csharp
We will eventually have non-UWP templates too (working on it as we speak), which are for win32 apps using WinUI 3 and the Windows App SDK (formerly known as Project Reunion)

packages/create-react-native-library/templates/example/example/windows/{%- project.name %}Example/App.xaml.cs
```diff
@@ -0,0 +1,49 @@
...
+
+#if DEBUG
+            InstanceSettings.UseDeveloperSupport = true;
+#else
+            InstanceSettings.UseDeveloperSupport = false;
+#endif
+
+            Microsoft.ReactNative.Managed.AutolinkedNativeModules.RegisterAutolinkedNativeModulePackages(PackageProviders); // Includes any autolinked modules
+
+            PackageProviders.Add(new Microsoft.ReactNative.Managed.ReactPackageProvider());
```
asklar, line 30: I think we recently deleted this line (this past week in fact :) )

packages/create-react-native-library/templates/example/example/windows/{%- project.name %}Example/Package.appxmanifest
```diff
@@ -0,0 +1,50 @@
...
+
+<Package
+  xmlns="http://schemas.microsoft.com/appx/manifest/foundation/windows10"
+  xmlns:mp="http://schemas.microsoft.com/appx/2014/phone/manifest"
+  xmlns:uap="http://schemas.microsoft.com/appx/manifest/uap/windows10"
+  IgnorableNamespaces="uap mp">
+
+  <Identity
+    Name="cb1b1ded-2f74-4bf4-89cd-ab411a2b6aa9"
+    Publisher="CN=Bartosz Klonowski"
```
asklar, line 11: this should probably be the user's name or something in the template :)

packages/create-react-native-library/templates/windows-cpp-library/windows/{%- project.name %}/readme.txt
```diff
@@ -0,0 +1,23 @@
+========================================================================
```
asklar, line 1: nit, can delete this file

packages/create-react-native-library/templates/windows-cpp-view-library/windows/{%- project.name %}/readme.txt
```diff
@@ -0,0 +1,23 @@
+========================================================================
```
asklar, line 1: can delete and remove this file from the project too :)

packages/create-react-native-library/templates/windows-cpp-view-library/windows/{%- project.name %}/{%- project.name %}ViewManager.hpp
```diff
@@ -0,0 +1,23 @@
...
+
+#include <winrt/Windows.UI.Xaml.Controls.h>
+#include <winrt/Windows.UI.Xaml.h>
+#include <winrt/Microsoft.ReactNative.h>
+
+namespace winrt::<%- project.name %>::implementation
+{
+    class <%- project.name %>ViewManager : public implements<
+        <%- project.name %>ViewManager,
+        Microsoft::ReactNative::IViewManager>
```
asklar, line 11: I think you'll probably want to implement some of the other interfaces, e.g. [IViewManagerWithReactContext](https://microsoft.github.io/react-native-windows/docs/IViewManagerWithReactContext)

asklar (approved):
Recommend adding CI validation that the apps/libs/modules can be generated properly as well as that they build :)

packages/create-react-native-library/templates/windows-cpp-view-library/windows/{%- project.name %}/{%- project.name %}ViewManager.hpp
BartoszKlonowski: I wanted to keep the starting point simple so a developer can add whatever will be used, but if you recommend adding this then I'll go with your advice 👍

packages/create-react-native-library/templates/example/example/windows/{%- project.name %}Example/Package.appxmanifest
BartoszKlonowski: I completely forgot about this, thanks!
I will double check whether there are more of such mistakes.

packages/create-react-native-library/templates/example/example/windows/{%- project.name %}Example/App.xaml.cs
BartoszKlonowski: This template is for late of v0.63, so I will need to keep this line, but once the whole Bob is updated, I'll go with removing it 👍

jahead:
What's the state of this?

BartoszKlonowski:
@jahead We are still waiting for the review and will only merge this with an approval of an owner: @satya164

jahead:
@satya164 do you have some time for you to review this?

satya164:
@BartoszKlonowski can you rebase this PR on top of main if you get some time this week? then I'll have time to review and merge this.

rtorr:
looking forward to this

satya164:
Closing due to inactivity. I'd like to figure out a different way to support other platforms without adding the code to this repo.

PR #174, opened 2021-05-28
remove buildToolsVersion

dulmandakh:
### Summary

Setting buildToolsVersion is no longer recommended, it'll default to Android Gradle Plugin defaults. It makes DX worse by requiring multiple versions of build tools by various dependencies require various versions.

### DIFF
--- a/packages/create-react-native-library/templates/example/example/android/build.gradle
+++ b/packages/create-react-native-library/templates/example/example/android/build.gradle
@@ -2,7 +2,6 @@
 
 buildscript {
     ext {
-        buildToolsVersion = "29.0.2"
         minSdkVersion = 16
         compileSdkVersion = 29
         targetSdkVersion = 29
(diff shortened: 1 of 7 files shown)


BartoszKlonowski (commented):
Indeed, we are using 3.2.1 and 3.5.3 for Java and Kotlin.
Looking at the changes I can see that all `buildToolsVersion`s in templates are covered, so it looks good in general, thank you!

---

But what about the examples?
https://github.com/callstack/react-native-builder-bob/blob/a78b9d147a2a2d1c42f050c60f808855bb97bc24/packages/create-react-native-library/templates/example/example/android/build.gradle#L5

Krizzu (approved):

BartoszKlonowski:
@dulmandakh My question about example was just a minor remark, so I've allowed myself to push to cover that.
Thank you once again for contribution!

dulmandakh:
@BartoszKlonowski I left example untouched because some third party modules still try to get it from root project. But it should be safe because most example app won't install third party modules. Sorry for late answer.

zamotany:
IMO this PR should have been merged only after the feedback from project owner.

Issue #191, opened 2021-08-28
expo sdk version not working with react-native-reanimated 2


TheEhsanSarshar:
The current expo SDK version in the example directory is not working with react-native-reanimated 2.
upgrade it to version 42

PR #192, opened 2021-08-28
update expo sdk version to 42

TheEhsanSarshar:
### Summary

The current SDK version (40) is not working with `react-native-reanimated 2`
so this PR will solve it.
fixing #191


satya164 (approved):

satya164:
Thanks!

PR #194, opened 2021-09-06
feat(bob): add `concurrent` and `failFast` config options

LukeeeeBennett:
#### `concurrent`

Set to `true` to run all build stages concurrently.

#### `failFast`

If the `concurrent` option is `true`, setting this option to `true` will exit all builds immediately if any build fails. Ignored if `concurrent` is `false`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,6 +64,8 @@ To configure your project manually, follow these steps:
    "react-native-builder-bob": {
      "source": "src",
      "output": "lib",
+     "concurrent": false,
+     "failFast": false,
      "targets": [
        ["aar", {"reverseJetify": true}],
        ["commonjs", {"copyFlow": true}],
@@ -185,6 +187,14 @@ Example:
 ```json
 ["aar", { "reverseJetify": true }]
 ```
+
+#### `concurrent`
+
+Set to `true` to run all build stages concurrently.
+
+#### `failFast`
+
+If the `concurrent` option is `true`, setting this option to `true` will exit all builds immediately if any build fails. Ignored if `concurrent` is `false`.
 
 ## FAQ
 
--- a/packages/react-native-builder-bob/src/index.ts
+++ b/packages/react-native-builder-bob/src/index.ts
@@ -377,6 +377,8 @@ yargs
       success: logger.success,
     };
 
+    let builds = [];
+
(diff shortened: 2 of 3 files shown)


lbennett-simplestream (approved):

packages/react-native-builder-bob/src/index.ts
```diff
@@ -385,45 +387,68 @@ yargs
...
     }
+
+    if (options.concurrent) {
+      const promises = builds.map((build) => build());
+      await (options.failFast
+        ? Promise.all(promises)
+        : Promise.allSettled(promises));
+    } else {
+      if (options.failFast) {
+        logger.warn('Concurrency is disabled. Ignoring failFast.');
```
Krizzu, line 445: Great 👍🏼

Krizzu (approved):
Great idea to speed those builds 👍🏼

README.md
```diff
@@ -186,6 +188,14 @@ Example:
...
 ```
 
+#### `concurrent`
+
+Set to `true` to run all build stages concurrently.
+
+#### `failFast`
+
+If the `concurrent` option is `true`, setting this option to `true` will exit all builds immediately if any build fails. Ignored if `concurrent` is `false`.
+
```
satya164, line 198: Is there any reason to provide options for these instead of making the builds work that way by default?

packages/react-native-builder-bob/src/index.ts
```diff
@@ -385,45 +387,68 @@ yargs
...
         default:
           logger.exit(`Invalid target ${chalk.blue(targetName)}.`);
       }
     }
+
+    if (options.concurrent) {
+      const promises = builds.map((build) => build());
+      await (options.failFast
+        ? Promise.all(promises)
+        : Promise.allSettled(promises));
```
satya164, line 442: I assume the logs will be out of order this way. Can you verify if the logs are still working correctly and properly grouped?

satya164 (changes requested):

satya164:
Hello @LukeeeeBennett, lemme know what you think about my comments

satya164:
Closing due to no reponse. Please open a new Pr if you still want to get this merged.

PR #195, opened 2021-09-06
Watch

LukeeeeBennett:
Add `watch` command.

I recommend merging https://github.com/callstack/react-native-builder-bob/pull/194 first


satya164:
See https://github.com/callstack/react-native-builder-bob/pull/194#issuecomment-1288142534

PR #198, opened 2021-09-16
fix(android): prefer downloading dependencies from Maven Central

troZee:
### Summary

Recent react-native dependencies are published to Maven Central only.

### Test plan

Check if `mavenCentral` is above `jcenter` in repo structure 

More context here: https://github.com/react-native-camera/react-native-camera/issues/2490

### DIFF
--- a/packages/create-react-native-library/templates/example/example/android/build.gradle
+++ b/packages/create-react-native-library/templates/example/example/android/build.gradle
@@ -8,6 +8,7 @@ buildscript {
     }
     repositories {
         google()
+        mavenCentral()
         jcenter()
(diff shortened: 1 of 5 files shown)


Krizzu (approved):

satya164 (approved):
Do we need to keep jcenter?

troZee:
> Do we need to keep jcenter?

I think, we can completely remove it in the future, since it has been removed in RN 65: https://react-native-community.github.io/upgrade-helper/?from=0.64.2&to=0.65.1 . For now, we can keep it as it is.

PR #199, opened 2021-09-25
fix: fix generating cpp project on iOS

satya164:
Currently, the CPP template doesn't import the header for the C code on iOS, so it fails to build. Additionally, it includes extraneous `AppDelegate.m` file which is not used.


packages/create-react-native-library/templates/cpp-library/ios/{%- project.name %}.mm
```diff
@@ -1,4 +1,5 @@
 #import "<%- project.name %>.h"
+#import "<%- project.identifier %>.h"
```
Krizzu, line 2: 👍🏼

Krizzu (approved):

PR #202, opened 2021-09-28
feat: print an error when library is not linked

satya164:
We should show better error messages when the library is not linked than the cryptic `"BlaBlaView" was not found in the UIManager` or `Native module cannot be null` error messages.

Screenshot_20210928-022628


Krizzu (approved):
👏

PR #220, opened 2022-01-11
fix: remove double-quoted imports from React-Core

Kudo:
### Summary

according to https://github.com/expo/expo/issues/15622#issuecomment-997225774, we should use `#import <React/RCTViewManager.h>` instead of double-quoted import.

### Test plan

regression test by creating a test library and running the example (`yarn && yarn example ios`).


vonovak (approved):

Kudo:
thank you, vonovak!

PR #246, opened 2022-05-28
fix: include header with project identifier for cpp

hyochan:
### Summary

The `cpp` template has the wrong include header.

### Test plan

```
npx create-react-native-library react-native-simple-turbo
cd react-native-simple-turbo
yarn example android
```


satya164 (approved):
Thank you

PR #275, opened 2022-09-08
chore: add new github workflow to test every template

atlj:
### Summary

With this GitHub workflow, we are now available to run `yarn lint`, `yarn test`, and `yarn typescript` on every possible example app

### Test plan

- [x] Turbo module with backward compat (experimental)
- [x] Turbo module (experimental)
- [x]  Native module
- [x]  Native view
- [x] JavaScript library
- [x] Expo app

### Future Plans

We might also test if the app builds for `ios` and `android` in the future.

### DIFF
--- a/.github/workflows/test-all-templates.yml
+++ b/.github/workflows/test-all-templates.yml
@@ -1,0 +1,63 @@
+name: Test all templates
+on:
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  test-all-templates:
+    name: Test all templates
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        example: [native]
+        type: [module, view]
+        language: [java-objc, java-swift, kotlin-objc, kotlin-swift, cpp, js]
+        include:
+          - language: js
+            example: expo
+            type: module
+    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/test-all-templates.yml
```diff
@@ -0,0 +1,63 @@
...
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        example: [native]
+        type: [module, view]
+        language: [java-objc, java-swift, kotlin-objc, kotlin-swift, cpp, js]
+        include:
+          - language: js
+            example: expo
+            type: module
```
MateWW, line 19: Supper minor thing but maybe we can keep one order of those keys.

MateWW (approved):
Looks great!

.github/workflows/test-all-templates.yml
atlj: Good idea, what is your preferred order I think
type
example
language

would be nice, I will also refactor the order I give these arguments

atlj:
This was closed because we created #276 with a new branch located from this repo instead of a fork

PR #273, opened 2022-09-06
feat: add fabric option

troZee:
### Summary

Close https://github.com/callstack/react-native-builder-bob/issues/238 

Screenshot 2022-09-06 at 15 25 48

### Test plan

- [ ] Turbo module with backward compat (experimental)
- [ ] Turbo module (experimental)
- [ ]  Native module
- [ ]  Native view
- [ ] JavaScript library

PR #276, opened 2022-09-09
chore: add new github workflow to test every template

atlj:
### Summary

With this GitHub workflow, we are now available to run `yarn lint`, `yarn test`, and `yarn typescript` on every possible example app

### Test plan

- [x] Turbo module with backward compat (experimental)
- [x] Turbo module (experimental)
- [x]  Native module
- [x]  Native view
- [x] JavaScript library
- [x] Expo app

### Future Plans

We might also test if the app builds for `ios` and `android` in the future.


atlj:
I don't know if we should fix existing lint problems in this PR or create a separate pr after #273 gets merged

satya164:
@atlj let's fix the lint problems here. we wouldn't want CI to be failing on main branch

satya164 (approved):

PR #326, opened 2022-11-15
refactor: use GitHub actions instead of CircleCI for generated project 

satya164:
This adds config for GitHub actions and removes the CircleCI config from the
generated project.

Previously, we kept CircleCI config as it provided better caching and was more
optimized due to multiple jobs being able to reuse the result from the one job,
reducing the setup time.

While the problems with setup time still remains with GitHub actions, the caching
part is better now which allows us to skip installing dependencies - reducing
those concerns. The biggest advantage of GitHub actions is that since it's
integrated with GitHub, the CI doesn't require additional steps after creating a
project.

In addition, we've moved to GitHub actions for this monorepo already, so it's
easier for us to maintain the config for GitHub actions as opposed a CircleCI
config that we don't use.


.github/actions/setup/action.yml
```diff
@@ -0,0 +1,25 @@
+name: Setup
+description: Setup Node.js and install dependencies
+
+runs:
+  using: composite
+  steps:
+    - name: Setup Node.js
+      uses: actions/setup-node@v3
+      with:
+        node-version: 16.x
```
thymikee, line 10: 18.x is not LTS, worth changing?

satya164: @thymikee in react navigation repo we had some issues with Node 18 and SSL changes so didn't try in this PR. will check if we can upgrade after this one is merged

packages/create-react-native-library/templates/common/$.github/workflows/ci.yml
```diff
@@ -0,0 +1,43 @@
...
+        uses: actions/checkout@v3
+
+      - name: Setup
+        uses: ./.github/actions/setup
+
+      - name: Lint files
+        run: yarn lint
+
+      - name: Typecheck files
+        run: yarn typescript
```
thymikee, line 19: Putting lint and typecheck in one job is intended? It used to be separate. However, libraries are small, so I'd expect this to be actually faster than spinning up the whole env twice to run them separately

thymikee (approved):

packages/create-react-native-library/templates/common/$.github/workflows/ci.yml
satya164: @thymikee yeah just to reduce setting up another job since it needs to clone repo and install deps all over again unlike circle ci

.github/workflows/build-templates.yml
```diff
@@ -110,6 +94,8 @@ jobs:
           path: |
             ${{ env.work_dir }}/**/node_modules
           key: ${{ runner.os }}-library-yarn-${{ hashFiles(format('{0}/**/package.json', env.work_dir)) }}
+          restore-keys: |
+            ${{ runner.os }}-library-yarn-
```
MateWW, line 98: Is this `-` at the end necessary?

satya164: @MateWW not in this specific case, but generally it's good to add it to avoid matching words that start with these letters so I generally follow this practice

MateWW: Oki, Thanks for explaination!

PR #350, opened 2023-01-16
feat(ts): new option to merge TS  declarations

wcandillon:
### Summary

The goal of this PR is to provide a new option named `mergeTSDeclarations` that merges type declarations within the `commonjs` and `module` targets. We are facing an issue where having the type declarations in a separate folder prevents us to do "custom" imports of a module. In the case of [React Native Skia](https://github.com/shopify/react-native-skia), for instance, we cannot import `@shopify/react-native-skia` on non-native environments directly. 
If we perform a "custom" import such as `@shopify/react-native-skia/lib/module/web` we don't have a type definition for it. We believe other React Native modules suffer from this issue as well.

But first things first, do you think such an option is a good idea?

We implemented our own custom merging [here](https://github.com/Shopify/react-native-skia/pull/1280).

### Test plan

Build package and test against this [PR](https://github.com/Shopify/react-native-skia/pull/1280).

### DIFF
--- a/packages/react-native-builder-bob/package.json
+++ b/packages/react-native-builder-bob/package.json
@@ -52,6 +52,7 @@
     "is-git-dirty": "^2.0.1",
     "json5": "^2.2.1",
     "kleur": "^4.1.4",
+    "merge-dirs": "^0.2.1",
     "prompts": "^2.4.2",
     "which": "^2.0.2",
     "yargs": "^17.5.1"
--- a/packages/react-native-builder-bob/src/index.ts
+++ b/packages/react-native-builder-bob/src/index.ts
@@ -12,6 +12,7 @@ import buildCommonJS from './targets/commonjs';
 import buildModule from './targets/module';
 import buildTypescript from './targets/typescript';
 import type { Options } from './types';
+import mergedirs from 'merge-dirs';
 
 // eslint-disable-next-line import/no-commonjs
 const { name, version } = require('../package.json');
@@ -424,6 +425,21 @@ yargs
           break;
         default:
           logger.exit(`Invalid target ${kleur.blue(targetName)}.`);
+      }
+    }
+    if (options.mergeTSDeclarations) {
+      const nonTSTargets = options.targets!.filter(
+        (target) => target !== 'typescript'
+      );
+      const tsTarget = path.resolve(
+        root,
+        output as string,
+        'typescript',
+        source as string
+      );
+      for (const target of nonTSTargets) {
+        const targetName = Array.isArray(target) ? target[0] : target;
+        mergedirs(tsTarget, path.resolve(root, output as string, targetName));
       }
     }
   })
--- a/packages/react-native-builder-bob/src/types.ts
+++ b/packages/react-native-builder-bob/src/types.ts
@@ -19,4 +19,5 @@ export type Options = {
   source?: string;
   output?: string;
   targets?: (Target | [Target, object])[];
+  mergeTSDeclarations?: boolean;
 };
--- a/yarn.lock
+++ b/yarn.lock
@@ -2167,6 +2167,11 @@ ajv@^6.10.0, ajv@^6.12.4, ajv@^6.12.6:
     json-schema-traverse "^0.4.1"
     uri-js "^4.2.2"
 
+ansi-escapes@^1.1.0:
+  version "1.4.0"
+  resolved "https://registry.yarnpkg.com/ansi-escapes/-/ansi-escapes-1.4.0.tgz#d3a8a83b319aa67793662b13e761c7911422306e"
+  integrity sha512-wiXutNjDUlNEDWHcYH3jtZUhd3c4/VojassD8zHdHCY13xbZy2XbW+NKQwA0tWGBVzDA9qEzYwfoSsWmviidhw==
+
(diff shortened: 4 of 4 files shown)


satya164:
Thanks for the PR! There are a couple of problems with this approach I think:

- The `package.json` entries are configured to have different files for different situations, e.g. when running in Node (including Jest & SSR), users need to include the `commonjs` files (to avoid build step), and `module` files for bundlers for treeshaking, etc. If users import from `commonjs`/`modules` directly, they'd either lose treeshaking on the client or need to configure to compile modules on the server.
- If the user mixes imports from `commonjs/`modules`/`src` in the same runtime then they'll be different instances of the module and can cause bugs depending on the library
- Importing directly from `lib` means that you have no control over what's internal and what's not since all your files are now part of external API and any change to file structure is a breaking change.

Usually, for this scenario, I'd create top-level folders with `package.json` files for this:

**web/package.json**:
```json
{
  "main": "../lib/commonjs/web/index.js",
  "module": "../lib/module/web/index.js",
  "react-native": "../src/web/index.ts",
  "types": "../lib/typescript/web/index.d.ts",
} 
```

Then user could import `@shopify/react-native-skia/web`. It can be a bit annoying when there are many folders like this but it avoids all issues of importing from `lib/` and could be automated as well.

wcandillon:
Based on the answer above, it looks like we have different expectation about what this PR does, could it be the case?

Copying the ts type declarations next to the module generated and the commonjs generated files is not expected to trigger any side-effects.

In the approach mentionned above, `"types": "../lib/typescript/web/index.d.ts"` is not expected to work.

wcandillon:
The original idea was to offer something symmetric to `copyFlow` but for typescript. But maybe the it is prefered to generate the commonjs and the module target directly with typescript (using a different tsconfig file for each target).

The approach used in this PR works great for us to run RN Skia on both Web and Node.js on top of the native target but maybe there are some details that are being overlooked to make this generally available.

satya164:
> Based on the answer above, it looks like we have different expectation about what this PR does, could it be the case?
> Copying the ts type declarations next to the module generated and the commonjs generated files is not expected to trigger any side-effects.

I'm not talking about what the PR does, it's about why it does that. The problems I mentioned are caused from importing code from `lib` folder that has different copies of the module. Not because you're copying the type definitions.

Since copying the type definitions this way is only necessary if someone imports from `lib` directly, I think it's important to understand why that pattern can cause issues.

> In the approach mentionned above, "types": "../lib/typescript/web/index.d.ts" is not expected to work.

Why is it not expected to work? We do this for `native-stack` for `react-native-screens` and it works fine. https://github.com/software-mansion/react-native-screens/blob/main/native-stack/package.json

> The original idea was to offer something symmetric to copyFlow but for typescript

The `copyFlow` approach is not ideal, it's implemented like that since that was the only way for Flow to find the type definitions, or at least as far as I was aware at the time I implemented it.

> But maybe the it is prefered to generate the commonjs and the module target directly with typescript

The same problems remain if you generate different targets with TypeScript. It's not about the tool. It's about having multiple targets - and hence multiple copies of the module.

Files from `lib/module` can't be imported on Node without a compilation step, but needed for treeshaking on client.

Files from `lib/commonjs` doesn't work with treeshaking on client, but can be imported on Node without a compilation step.

Mixing files from `lib/module` and `lib/commonjs` in the same project may cause bugs due to different instances of the module - if there are any global variables, side effects etc.

> The approach used in this PR works great for us to run RN Skia on both Web and Node.js on top of the native target

I'm not sure what you mean by "native target", but if it works on Node.js then you likely have a compilation step somewhere. Next.js has a compilation step for server-side code, Jest can be configured to compile specific modules, etc.

While it may work in your specific case, it's not suitable for generic usage due to the issues I mentioned.

If you need clarification on any of the issues I mentioned then I'm happy to elaborate.

wcandillon:
I haven't played with it recently but last-time I checked react-navigation suffers from the same issue where using a private API (via "custom" import) would result of the type definition not being found (despite the type declaration existing in the package). This private API where pulled to implement custom navigators and it was well understood that these are private/non-backward compatible APIs but it got the job done.

In the case of React Native Skia Web, we cannot import `@shopify/react-native-skia` before a web assembly file is loaded asynchronously via `@shopify/react-native-skia/lib/module/web`. 

On Node.js same story but we use the commonjs version of the module `@shopify/react-native-skia/lib/commonjs/node`.

I'm closing this PR since there doesn't seem to be any appetite for it.

satya164:
> I haven't played with it recently but last-time I checked react-navigation suffers from the same issue where using a private API 

It's not really an issue in the case of React Navigation since that's not a supported use case. People shouldn't be doing it - it can easily cause issues since React Native apps need to import `src` while jest uses `lib`. If there is anything that needs to be exported for custom navigators, we usually do export them.

> On Node.js same story but we use the commonjs version of the module @shopify/react-native-skia/lib/commonjs/node.

For the common use case, file running in Node imports client files, client files import from `lib/module` - hence the Node imports `lib/module` as well. There's no conditional import to be done for `commonjs`/`module`. This is applicable for SSR, Jest etc.

> I'm closing this PR since there doesn't seem to be any appetite for it.

It's sad to hear this after I tried to explain the current problems with this approach and offer suggestions.

PR #354, opened 2023-01-27
feat: make year dynamic

dcangulo:
### Summary
We are manually updating year when it can be dynamic.

### Test plan
* The following command should not fail.
```
node packages/create-react-native-library/bin/create-react-native-library sample-module
```
* Year in LICENSE should be the device's year.

### DIFF
--- a/packages/create-react-native-library/src/index.ts
+++ b/packages/create-react-native-library/src/index.ts
@@ -577,6 +577,7 @@ async function create(argv: yargs.Arguments<any>) {
     repo: repoUrl,
     example,
     turborepo,
+    year: new Date().getFullYear(),
   };
 
   const copyDir = async (source: string, dest: string) => {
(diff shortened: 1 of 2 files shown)


packages/create-react-native-library/src/index.ts
```diff
@@ -557,6 +557,7 @@ async function create(argv: yargs.Arguments<any>) {
             `${project.charAt(0).toUpperCase()}${project
               .replace(/[^a-z0-9](\w)/g, (_, $1) => $1.toUpperCase())
               .slice(1)}`,
+      year: new Date().getFullYear(),
```
atlj, line 560: It would make more sense if the `year` could be top-level instead of inside the `project`. Could you place the `year` directly inside the options?

Then you can use it as `<%- year -%>` in the template.

atlj (changes requested):
Hey! thank you for the PR left a small nit 🚀

dcangulo:
@atlj Updated

atlj (approved):
LGTM, thank you so much for the PR. 🚀🚀

atlj:
Please feel free to merge

PR #358, opened 2023-02-20
refactor: use install_modules_dependencies method only if react native version >=0.71.0

ykhateeb:
This pull request replaces all the inline IOS dependencies in **podspec** and uses `install_modules_dependencies` method to depend on React Native core changes.

### DIFF
--- a/.github/workflows/build-templates.yml
+++ b/.github/workflows/build-templates.yml
@@ -103,6 +103,7 @@ jobs:
 
       - name: Create library
         run: |
+          rm -rf ${{ env.work_dir }} # Workaround for tests failing intermittently
           ./packages/create-react-native-library/bin/create-react-native-library ${{ env.work_dir }} \
             --slug @bob/react-native-test \
             --description test \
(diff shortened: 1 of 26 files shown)


satya164 (commented):
Thank you for the PR. Can you provide some more details about this method like what does this do, why was this added, which version of React Native does this support etc.?

ykhateeb:
Hey @satya164,

This method contains all the needed IOS dependencies (same as the current that we have now) with supporting old and new architecture, the purpose of adding it is to simplify the process of migrating to a new React Native version, by depending on React Native changes rather than duplicate same code, this method exists starting from  v0.71.0.

Here is the full implementation: https://github.com/facebook/react-native/blob/febf6b7f33fdb4904669f99d795eba4c0f95d7bf/scripts/cocoapods/new_architecture.rb#L79

satya164 (changes requested):
@ykhateeb thanks. I don't want to drop support for older RN versions yet until 0.71 is a few releases old. Can you wrap this in the condition so that this is only used for new architecture?

ykhateeb:
@satya164  Could please check it now?

PR #365, opened 2023-02-28
refactor: remove unnecessary babel-plugin-module-resolver

Sunbreak:
### Summary

As far as I test, `babel-plugin-module-resolver` has been unnecessary since react-native@0.69.4

And it caused some problem when migerating some legacy library, e.g. https://github.com/react-native-linear-gradient/react-native-linear-gradient

### Test plan

### DIFF
--- a/packages/create-react-native-library/src/utils/generateExampleApp.ts
+++ b/packages/create-react-native-library/src/utils/generateExampleApp.ts
@@ -33,9 +33,7 @@ const PACKAGES_TO_REMOVE = [
   'typescript',
 ];
 
-const PACKAGES_TO_ADD_DEV = {
-  'babel-plugin-module-resolver': '^4.1.0',
-};
+const PACKAGES_TO_ADD_DEV = {};
 
 const PACKAGES_TO_ADD_WEB = {
   'react-dom': '18.1.0',
--- a/packages/create-react-native-library/templates/example-legacy/example/babel.config.js
+++ b/packages/create-react-native-library/templates/example-legacy/example/babel.config.js
@@ -1,17 +1,3 @@
-const path = require('path');
-const pak = require('../package.json');
-
 module.exports = {
   presets: ['module:metro-react-native-babel-preset'],
-  plugins: [
-    [
-      'module-resolver',
-      {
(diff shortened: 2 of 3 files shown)


Sunbreak:
When applying `babel-plugin-module-resolver` onto `react-native-linear-gradient` at https://github.com/Sunbreak/react-native-linear-gradient/commit/cb4236380b56dff7539255f36be1bdea2ebc974f

Screenshot_20230228_101658

```
SyntaxError: "index.js: The "path" argument must be of type string. Received undefined"
    ...
    Object.<anonymouse>(/Users/sunbreak/w/Sunbreak/react-native-linear-gradient/Examples/FabricExample/babel.config.js:12.28
    ...
```

Which is

```js
...
          [pak.name]: path.join(__dirname, '../..', pak.source),
...
```

Sunbreak:
And https://github.com/Sunbreak/react-native-linear-gradient/commit/32a58c183dc43ab12130394dbf0512b9de8a7bba fixed

Check the PR for more detail: https://github.com/react-native-linear-gradient/react-native-linear-gradient/pull/586

satya164:
Hey @Sunbreak, thanks for the PR. Can you provide more details on why the babel plugin isn't needed from react-native 0.69 onwards?

The error you shared looks specific to your project where it's missing `source` field in `package.json`, but using it in `babel.config.js`.

Sunbreak:
> Hey @Sunbreak, thanks for the PR. Can you provide more details on why the babel plugin isn't needed from react-native 0.69 onwards?
> 
> The error you shared looks specific to your project where it's missing `source` field in `package.json`, but using it in `babel.config.js` resulting in `undefined` passed to `path.join`.

I'm not sure whether `soruce` is necessary. Maybe we could check `source` before using it?

satya164:
It's not necessary, but it's a convention that we use in the default template. It won't make sense to check for this given that it'll always be present in the template. If your project has different setup then the config should be adjusted in the project - e.g. either add source property or refer to the path directly instead of reading from the property.

Sunbreak:
Got it. Closed by design

PR #383, opened 2023-04-26
fix: Error when creating a package. Error in step `Copying template` …

### LINKED ISSUE #382: Error when creating a package. Error in step `Copying template`
### Description

The first time I try this repo.

When running `npx create-react-native-library@latest name` and fill all the `package.json` data, this happens:
imatge

```bash
/Users/aaaa/.npm/_npx/e106e660d51959d2/node_modules/create-react-native-library/lib/utils/generateExampleApp.js:62
        reject(new Error(stderr));
               ^

Error: - Downloading template
✔ Downloading template
- Copying template
✖ Copying template
error Error: Couldn't find the "/var/folders/ys/spt316fs5sxck01gjy40s1rc0000gn/T/rncli-init-template-AoLPqU/node_modules/react-native/template.config.js file inside "react-native" template. Please make sure the template is valid. Read more: https://github.com/react-native-community/cli/blob/master/docs/init.md#creating-custom-template.

    at ChildProcess.<anonymous> (/Users/aaaa/.npm/_npx/e106e660d51959d2/node_modules/create-react-native-library/lib/utils/generateExampleApp.js:62:16)
    at Object.onceWrapper (node:events:642:26)
    at ChildProcess.emit (node:events:527:28)
    at Process.ChildProcess._handle.onexit (node:internal/child_process:291:12
```

The line 62 of `generateExampleApp.js` in the npx's cache `~/.npm/_npx/e106e660d51959d2/node_modules/create-react-native-library/lib/utils`

```js
        reject(new Error(stderr));
```

I tried version `0.27.2`. The same problem.

### Packages

- [X] create-react-native-library
- [ ] react-native-builder-bob

### Selected options

The step `What type of library do you want to develop?` I tried all, only `Javascript` worked.

### Link to repro

_No response_

### Environment

- macOS 13
- Node 16.15.1

dvlkv:
Solves #382

### Summary

Fixes #382 according to react-native [issue](https://github.com/facebook/react-native/issues/36290)

### Test plan

#382 won't reproduce with yarn installed.

### DIFF
--- a/packages/create-react-native-library/src/utils/generateExampleApp.ts
+++ b/packages/create-react-native-library/src/utils/generateExampleApp.ts
@@ -73,6 +73,7 @@ export default async function generateExampleApp({
           '--version',
           reactNativeVersion,
           '--skip-install',
+          '--npm'
         ]
       : // `npx create-expo-app example --no-install`
         ['create-expo-app@latest', directory, '--no-install'];
--- a/packages/create-react-native-library/src/utils/generateExampleApp.ts
+++ b/packages/create-react-native-library/src/utils/generateExampleApp.ts
@@ -73,7 +73,7 @@ export default async function generateExampleApp({
           '--version',
           reactNativeVersion,
           '--skip-install',
-          '--npm'
+          '--npm',
         ]
       : // `npx create-expo-app example --no-install`
         ['create-expo-app@latest', directory, '--no-install'];


atlj:
Thanks for the PR!

Although forcing the CLI to use `npm` seems to fix the issue, we would still like to use `yarn`. I have gone through many issues and saw some people can see this issue while others are not.  I've noticed newer versions of `yarn` such as `3.5.0` are causing this issue. I think until this issue is solved inside the CLI we can have this fix and remove it once it's fixed.

@satya164 WDTY?

dvlkv:
Another solution is to add a flag which disables yarn. People who facing this issue can use this flag until it is not resolved inside the CLI.

satya164:
It's important to note that Yarn is required in the default template. There are scripts, CI etc. that uses Yarn. So adding a flag to disable Yarn for creating project doesn't help much when users will need to use it after creating the project anyway.

Maybe we should detect and throw an error if Yarn isn't installed.

satya164:
Though I also don't understand this - if it cannot be reproduced when Yarn is installed, what difference does `--npm` make? Isn't `npm` used regardless since Yarn isn't installed?

atlj:
@satya164 the problem happens when the user has a newer version of yarn like `3.5.0`. The cli fails if that's the case. I've seen some discussion among cli folks to address this issue [source](https://github.com/react-native-community/cli/pull/1931).

Forcing cli to use npm if a newer version of yarn is installed might fix the issue. So maybe we can add a function to check the yarn version and if it's newer than a certain version, we can force cli init to use npm.

satya164:
@atlj gotcha. i was confused because the PR description says that it can't be reproduced when yarn is installed. i don't mind using `--npm` internally - but I'm against adding a flag to disabled yarn.

We can internally use an workaround but it probably makes more sense to fix it in CLI instead. But if it'll take too long or is too complicated, we can use `--npm` internally. We'll need to ensure the final project doesn't include any npm specific stuff like `package-lock.json`.

atlj:
@satya164 I've just checked the PR and no `package-lock.json` is generated since we pass the `--skip-install` flag to cli. I think it's good to merge but we should also watch this PR https://github.com/react-native-community/cli/pull/1931 to remove this fix.

atlj:
@dvlkv could you address the linting issue (you need to add `,`) please see: https://github.com/callstack/react-native-builder-bob/actions/runs/4808998616/jobs/8632552854?pr=383

packages/create-react-native-library/src/utils/generateExampleApp.ts
```diff
@@ -73,6 +73,7 @@ export default async function generateExampleApp({
           '--version',
           reactNativeVersion,
           '--skip-install',
+          '--npm'
```
atlj, line 76: Could you address the linting issue here? (add a `,`)

atlj (changes requested):
Please address the linting issue

atlj (changes requested):
Please address the linting issue

packages/create-react-native-library/src/utils/generateExampleApp.ts
dvlkv: Done!

guillempuche:
Could you release the new version?

PR #435, opened 2023-08-05
feat(builder-bob): dont use npm bin to get tsc path

### LINKED ISSUE #434: Failed to locate 'tsc' in workspace with npm / pnpm
### Description

Hi! I'm seeing the following error when running `bob build`:

```
Failed to locate 'tsc' in the workspace. Falling back to the globally installed version. Consider adding typescript to your devDependencies or specifying the tsc option for the typescript target.
```

I think this is only an issue when using `npm` or `pnpm` instead of `yarn`. In my case I don't have a choice because I'm working in a `pnpm` monorepo 😄 

I looked at the code and in `lib/targets/typescript.js`, it attempts to run `npm bin` to get the path to the `bin` folder.
When I log the response of this, it appears that the `npm bin` command doesn't exist. I'm on npm v9.5.1.

I can't find any release notes around `npm bin` being removed, but strangely I don't see it in their docs anywhere for v9. It seems to have just disappeared after v8: https://docs.npmjs.com/cli/v9/commands.

For anyone else that runs into this: a workaround is to explicitly add the path to the `tsc` command in the `targets` config:

```
"react-native-builder-bob": {
    "source": "src",
    "output": "lib",
    "targets": [
      "commonjs",
      "module",
      [
        "typescript",
        {
          "project": "tsconfig.json",
          "tsc": "./node_modules/.bin/tsc" // here!!
        }
      ]
    ]
  },
```

You can reproduce this with a pretty standard setup:

- Run `npx create-react-native-library@latest some-cool-name`
- Select `JavaScript library`
- Run `yarn` or `npm install` in the generated project (which one you run doesn't seem to matter)
- Run `npm run prepack`

### Packages

- [ ] create-react-native-library
- [X] react-native-builder-bob

### Selected options

Basic configuration via `npx create-react-native-library@latest`. Selected `Javascript Library`. Nothing else changed

### Link to repro

_No response_

### Environment

```
System:
  OS: macOS 12.6.8
  CPU: (10) arm64 Apple M1 Pro
  Memory: 108.08 MB / 32.00 GB
  Shell:
    version: 5.8.1
    path: /bin/zsh
Binaries:
  Node:
    version: 18.16.0
    path: ~/.nvm/versions/node/v18.16.0/bin/node
  Yarn:
    version: 1.22.19
    path: ~/.nvm/versions/node/v18.16.0/bin/yarn
  npm:
    version: 9.5.1
    path: ~/.nvm/versions/node/v18.16.0/bin/npm
  Watchman:
    version: 2023.05.22.00
    path: /opt/homebrew/bin/watchman
Managers:
  CocoaPods:
    version: 1.12.1
    path: /Users/jon.rigert/.rbenv/shims/pod
SDKs:
  iOS SDK:
    Platforms:
      - DriverKit 22.1
      - iOS 16.1
      - macOS 13.0
      - tvOS 16.1
      - watchOS 9.1
  Android SDK: Not Found
IDEs:
  Android Studio: 2021.2 AI-212.5712.43.2112.8815526
  Xcode:
    version: 14.1/14B47b
    path: /usr/bin/xcodebuild
Languages:
  Java:
    version: 11.0.16
    path: /usr/bin/javac
  Ruby:
    version: 2.7.5
    path: /Users/jon.rigert/.rbenv/shims/ruby
npmPackages:
  "@react-native-community/cli": Not Found
  react:
    installed: 18.2.0
    wanted: 18.2.0
  react-native:
    installed: 0.72.3
    wanted: 0.72.3
  react-native-macos: Not Found
npmGlobalPackages:
  "*react-native*": Not Found
Android:
  hermesEnabled: Not found
  newArchEnabled: Not found
iOS:
  hermesEnabled: Not found
  newArchEnabled: Not found
```

atlj:
## Summary

Closes #434
The builder bob typescript build phase uses `yarn bin` and `npm bin` commands to locate the `tsc`. However, the `npm bin` command is removed with npm 9.0. Please see: https://github.blog/changelog/2022-10-24-npm-v9-0-0-released/

## Test plan

> Both of the test cases start with the following steps:
1. Clone the repo and cd into `packages/react-native-builder-bob`
2. Run `yarn prepack` to build the builder bob.
3. Run `npm link`
4. Create a new project with `npx create-react-native-library@latest my-awesome-project`. You can select `javascript` as the type.
5. Run `npm link react-native-builder-bob` to link the project.

### Resolve `tsc` automatically

1. Run `npm run prepack` to verify the typescript compilation step works properly.
2. Run `yarn prepack` to verify the typescript compilation step works properly.
3. Run `pnpm prepack` to verify the typescript compilation step works properly.

### Resolve `tsc` using the builder bob config

1. Go to the `package.json` file of the project you've generated.
2. Go to the `react-native-builder-bob` section and specify the tsc binary as following: (the `tsc` line needs to be added.)
```json
  "react-native-builder-bob": {
    "source": "src",
    "output": "lib",
    "targets": [
      "commonjs",
      "module",
      [
        "typescript",
        {
          "project": "tsconfig.build.json",
          "tsc": "./node_modules/.bin/tsc"
        }
      ]
    ]
  }
```
3. Run `npm run prepack` to verify the typescript compilation step works properly.
4. Run `yarn prepack` to verify the typescript compilation step works properly.
5. Run `pnpm prepack` to verify the typescript compilation step works properly.

### DIFF
--- a/packages/react-native-builder-bob/package.json
+++ b/packages/react-native-builder-bob/package.json
@@ -53,7 +53,6 @@
     "json5": "^2.2.1",
     "kleur": "^4.1.4",
     "prompts": "^2.4.2",
-    "which": "^2.0.2",
     "yargs": "^17.5.1"
   },
   "optionalDependencies": {
--- a/packages/react-native-builder-bob/package.json
+++ b/packages/react-native-builder-bob/package.json
@@ -53,6 +53,7 @@
     "json5": "^2.2.1",
     "kleur": "^4.1.4",
     "prompts": "^2.4.2",
+    "which": "^2.0.2",
     "yargs": "^17.5.1"
   },
   "optionalDependencies": {
--- a/packages/react-native-builder-bob/src/targets/typescript.ts
+++ b/packages/react-native-builder-bob/src/targets/typescript.ts
@@ -81,6 +81,16 @@ export default async function build({
 
     if (options?.tsc) {
       binary = path.resolve(root, options.tsc);
+
+      if (!(await fs.pathExists(binary))) {
+        throw new Error(
+          `The ${kleur.blue(
+            'tsc'
+          )} binary doesn't seem to be installed at ${kleur.blue(
+            binary
+          )}. Please specify the correct path in options or remove it to use the workspace's version.`
+        );
+      }
     } else {
       binary = 'npx';
     }
--- a/packages/react-native-builder-bob/src/targets/typescript.ts
+++ b/packages/react-native-builder-bob/src/targets/typescript.ts
@@ -1,11 +1,9 @@
 import kleur from 'kleur';
 import path from 'path';
 import fs from 'fs-extra';
-import which from 'which';
 import spawn from 'cross-spawn';
 import del from 'del';
 import JSON5 from 'json5';
-import { platform } from 'os';
 import type { Input } from '../types';
 
 type Options = Input & {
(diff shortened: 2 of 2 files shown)


satya164 (changes requested):
I think it's important to ensure that we're running the local version of `tsc`. Using a global version of `tsc` isn't desired and may produce in inconsistent output. That's why there was a warning previously.

Using `npx` like this is problematic as it will try to install the package if it's not installed (the `tsc` package isn't even the official one), or run the global version if it's not present locally. I'm also not sure how it'll work in yarn monorepo (or regular projects using yarn).

The test plan also mentions that `tsc` option needs to be specified in bob's config - so I'm not sure if `npx` is working here as specifying the config skips the logic that uses `npx`.

This change needs to keep the previous behavior:

- Make sure that it executes locally installed `tsc` bindary
- If it's not installed locally, either use global version with a warning, or throw an error
- Make sure that this change works in repos using yarn, pnpm, etc.,

atlj:
I have removed the `npx` logic and now just using a bare path to determine where tsc is.
This is due to two reasons:
1. `npm bin` is deprecated and [it is not aware of workspaces](https://docs.npmjs.com/cli/v8/commands/npm-bin?v=true#synopsis)
2. npm binaries are always linked under `node_modules/.bin`. Please see [this](https://docs.npmjs.com/cli/v9/configuring-npm/folders?v=true#executables).

>The test plan also mentions that tsc option needs to be specified in bob's config - so I'm not sure if npx is working here as specifying the config skips the logic that uses npx.

> Make sure that this change works in repos using yarn, pnpm, etc.,

I've updated the testing section to support multiple cases

satya164 (approved):

PR #442, opened 2023-08-17
Feature: Allow bob.config.cjs

samijaber:
### Summary

I have a library configured with `"type": "module"`. This means that I need to provide `bob.config.cjs`, as `.js` extensions won't do. That config file needs to be explicitly added to the cosmicConfig

### DIFF
--- a/packages/react-native-builder-bob/src/index.ts
+++ b/packages/react-native-builder-bob/src/index.ts
@@ -18,7 +18,7 @@ const { name, version } = require('../package.json');
 
 const root = process.cwd();
 const explorer = cosmiconfigSync(name, {
-  searchPlaces: ['package.json', `bob.config.js`],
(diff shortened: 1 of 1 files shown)


satya164 (commented):
Thanks for the PR. Is it possible to add support for `.mjs` as well? I think it'd make sense if we're supporting `.cjs` extension.

samijaber:
@satya164 I did try to add `.mjs` as well but could not get it working out-of-the-box. Don't remember exactly what happened, but I got some type of error from CosmicConfig not knowing how to load `.mjs` extensions.

Could we still merge this and explore adding `.mjs` in a separate PR/branch?

satya164:
@samijaber sounds good

satya164 (approved):

PR #450, opened 2023-08-24
feat(bob): add ignore pattern to config

atlj:
### Summary

This adds the possibility to pass custom ignore patterns while working with builder bob.

### Test plan

1. Create a new library using `create-react-native-library`. You can select `Native module`, and `Kotlin & Swift` options.
2. Go to `package.json`, and find the `react-native-builder-bob` section
3. Replace `commonjs` target as following:
```json
{
    "exclude": "**/ignore_me/**"
    "targets": [
      "commonjs",
      "module",
      [
        "typescript",
        {
          "project": "tsconfig.build.json"
        }
      ]
    ]
}
```
4. Create a new folder named `src/ignore_me` and put `.ts` files inside it
5. Build the library and make sure the files inside the `ignore_me` aren't included in the `commonjs` and `module` targets.


README.md
```diff
@@ -137,6 +137,20 @@ The options can be specified in the `package.json` file under the `react-native-
 
 The name of the folder with the source code which should be compiled. The folder should include an `index` file.
 
+#### `exclude`
```
satya164, line 140: Move it to after `output`, since the `source` and `output` options are related to each other and it's better to have them closer.

packages/react-native-builder-bob/src/index.ts
```diff
@@ -418,7 +418,7 @@ yargs
             root,
             source: path.resolve(root, source as string),
             output: path.resolve(root, output as string, 'commonjs'),
-            options: targetOptions,
+            options: { ...targetOptions, exclude: options.exclude },
```
satya164, line 421: We are use the pattern of passing top-level options in the argument object (like `source`, `output`) and target specific options in the `options` field. So lets do the same for `exclude`.

satya164 (changes requested):

satya164 (approved):

PR #470, opened 2023-09-29
refactor: remove the *.xcodeproj from ios template

satya164:
### Summary

This removes the `*.xcodeproj` from the `ios` directory in the generated library. These files are not needed for autolinking, and keeping them has few downsides:

- People open this project instead of example and it doesn't work, so it adds confusion
- It's a big unreadable file to maintain, both for library developers and for us

### Test plan

Generate a project without the `*.xcodeproj` and verify that it compiles and runs


atlj (approved):

YOEL311:
Hey @satya164 
I need to create a module that uses with local library ``.framework``
how i can link the ``.framework`` files without the ``xcodeproj`` file?

gausam:
@satya164 could this perhaps return as an optional flag during install? I've also run into the same need as @YOEL311's.

satya164:
@gausam dependencies in react native libraries are managed with cocoapods, not with xcodeproj file. you need to include your framework using cocoapods https://github.com/callstack/react-native-builder-bob/discussions/379

PR #478, opened 2023-10-06
[WIP] chore: comment diffs automatically

atlj:
Please don't mind this PR yet, this is just for testing purposes.


atlj:
PLZ also test no diff

.github/workflows/comment-diffs.yml
```diff
@@ -0,0 +1,306 @@
...
+              "module-new": "Turbo module",
+              "view-legacy": "Native view",
+              "view-mixed": "Fabric view with backward compat",
+              "view-new": "Fabric view",
+              "java-objc": "Java and Objective C",
+              "java-swift": "Java and Swift",
+              "kotlin-objc": "Kotlin and Objective C",
+              "kotlin-swift": "Kotlin and Swift",
+              ".txt": ""
+            }
```
satya164, line 273: having the matrix in JSON would be nice, since this table could also be colocated.

though maybe we should add a command to `create-react-native-library`, `create-react-native-library --list-types` and `create-react-native-library --list-languages` or something that lists all available templates instead hardcoding. then we could use that for matrix in build-templates as well as here. i have forgotten to change the matrix when i change templates in past (worse when something gets added and there's no test failure), and it'll avoid this.

.github/workflows/comment-diffs.yml
```diff
@@ -0,0 +1,306 @@
...
+
+            const parsedArtifacts = fs.readdirSync('./outputs').map(fileName => {
+              const [type, language] = fileName.split("+");
+
+              const title = Object.entries(replaceTable).reduce((acc, [key, value]) => {
+                return acc.replace(new RegExp(key, "g"), value);
```
satya164, line 279: wouldn't `return acc.replace(key, value);` work here?

Issue #13, opened 2019-12-17
Alternative for `prepare` script to work around https://github.com/yarnpkg/yarn/issues/7212


lfalkner:
Is there an alternative to where `bob build` can run that does not involve using the `prepare` script? 

There is a [known issue](https://github.com/yarnpkg/yarn/issues/7212) with `yarn` that causes a second `yarn` process to spin up when a git package contains a `prepare` script. This makes it difficult to contribute to open source for any package which has such a script.

If there is an alternative to when and where the `bob build` script should run, that would be ideal.

lfalkner:
Note this work-around worked for us: https://github.com/yarnpkg/yarn/issues/7212#issuecomment-487930971

Happy to contribute back a documentation update if you'd like

satya164:
Closing since we use `prepack` script now instead.

PR #497, opened 2023-11-17
fix: use prepack instead of prepare

pwltr:
### Summary

Since Yarn v2 and upwards [doesn't support `prepare` script](https://yarnpkg.com/advanced/lifecycle-scripts) anymore I believe `prepack` should be used in the template to run `bob build` on install.

[This comment](https://github.com/callstack/react-native-builder-bob/issues/13#issuecomment-1620011941) says this should already be set but it wasn't in my case.

### DIFF
--- a/packages/create-react-native-library/templates/common/$package.json
+++ b/packages/create-react-native-library/templates/common/$package.json
@@ -37,7 +37,7 @@
 <% } else { -%>
     "clean": "del-cli lib",
 <% } -%>
-    "prepare": "bob build",
+    "prepack": "bob build",
     "release": "release-it"
   },
   "keywords": [


satya164:
prepare is used because it runs when people install the package from git. yarn 3 not running it on install is not important since it's not necessary anyway.

pwltr:
> prepare is used because it runs when people install the package from git

Could you elaborate? I am running yarn v1 and `prepare` is not running so I'm trying to migrate to v4 and using `prepack`. How is transpiling source files on install not necessary with yarn 3+? What would be the canonical way to get the build files if not with prepare/prepack?

satya164:
> > yarn 3 not running it on install is not important since it's not necessary anyway
> 
> 
> 
> Could you elaborate? I am running yarn v1 and `prepare` is not running so I'm trying to migrate to v4 and using `prepack`. How is transpiling source files on install not necessary with yarn 3+? What would be the canonical way to get the build files if not with prepare/prepack?

Why is it necessary to compile files during install? Yarn 3 or otherwise. Compiled files are for consumers of the npm package and they only need to be compiled during publish step or when installing from Git/local directory.

pwltr:
I'm pulling the package from Github (no npm package yet) and using yarn v1 prepare/prepack is not running. Only working setup is with prepack and yarn v4 so far. Also tried `postinstall` but no luck either. I would like to stick with install the package from Github but it sounds like that is not a good option?

satya164:
https://callstack.github.io/react-native-builder-bob/build#manual-configuration

- `prepare` is run when the package is published, as well as when it's installed from a git URL. It may also run when dependencies are installed based on the package manager.
- `prepack` only runs when the package is packed for publishing.

You want `prepare` because that's run when people when publishing and when package is installed from Git repo with `npm`. `npm` doesn't run `prepack` script when installing from Git https://github.com/callstack/react-native-paper/issues/4082 so it makes it unusable. If Yarn doesn't run `prepare` on installing from Git then that's a Yarn issue.

Either way both `prepack` and `prepare` run when publishing the package as the default setup is configured for npm and npm is used to publish the library.

satya164:
I also tested a bit more and when installing from git:

- Yarn 1 and npm run only `prepare`
- Yarn 4 runs only `prepack`

So Yarn 4's behavior makes no sense and if the package is for general consumption, `prepare` is more appropriate.

pwltr:
Very much appreciated, will try again with `prepare` that would save me a ton of headaches.

PR #501, opened 2023-12-03
chore: use project identifier instead of generic

dcangulo:
### Summary

Doesn't really fix/improve anything, just a nitpick. `cpp` seems too generic. Just close if it's not a welcome change. Thanks!

### Test plan

```
yarn example android
```


satya164 (approved):
Thanks

PR #519, opened 2024-01-27
add overrides for npm

keremoge:
@types/react: add missing overrides for *npm*


satya164:
Thanks for the PR. However the default template doesn't support npm as mentioned in the `CONTRIBUTING.md` inside your project.

keremoge:
Thanks, we use npm with no issues 👍🏽

Issue #518, opened 2024-01-27
Error while installing pods with swift library


mattiaferrari02:
### Description

Create the library selecting the options to write the native module using swift code.
Type `yarn` to install all node modules
Go to `example/ios` type `pod install`
Error occurs saying
```
The Swift pod `react-native-my-library` depends upon `glog`, which does not define modules. 
To opt into those targets generating module maps (which is necessary to import them from Swift when building as static libraries),
 you may set `use_modular_headers!` globally in your Podfile, or specify `:modular_headers => true` for particular dependencies
```
If I `use_modular_headers` other error occurs.
What can i do?

### Packages

- [X] create-react-native-library
- [ ] react-native-builder-bob

### Selected options

Native Module > Java & Swift

### Link to repro

https://github.com/mattiaferrari02/test-lib-swift

### Environment

```
System:
  OS: macOS 12.6.8
  CPU: (8) arm64 Apple M1
  Memory: 115.11 MB / 16.00 GB
  Shell:
    version: 5.8.1
    path: /bin/zsh
Binaries:
  Node:
    version: 18.19.0
    path: ~/.nvm/versions/node/v18.19.0/bin/node
  Yarn:
    version: 3.6.1
    path: ~/.nvm/versions/node/v18.19.0/bin/yarn
  npm:
    version: 10.2.3
    path: ~/.nvm/versions/node/v18.19.0/bin/npm
  Watchman:
    version: 2023.03.13.00
    path: /opt/homebrew/bin/watchman
Managers:
  CocoaPods:
    version: 1.12.0
    path: /opt/homebrew/bin/pod
SDKs:
  iOS SDK:
    Platforms:
      - DriverKit 22.2
      - iOS 16.2
      - macOS 13.1
      - tvOS 16.1
      - watchOS 9.1
  Android SDK:
    API Levels:
      - "28"
      - "29"
      - "30"
      - "31"
      - "32"
      - "33"
    Build Tools:
      - 28.0.3
      - 29.0.2
      - 30.0.2
      - 30.0.3
      - 31.0.0
      - 33.0.0
    System Images:
      - android-29 | Google Play ARM 64 v8a
      - android-31 | Google APIs ARM 64 v8a
      - android-31 | Google APIs Intel x86 Atom_64
      - android-31 | Google Play ARM 64 v8a
    Android NDK: 21.4.7075529
IDEs:
  Android Studio: 2021.3 AI-213.7172.25.2113.9014738
  Xcode:
[...]

PR #523, opened 2024-02-02
chore: use the macOS 13 runner

atlj:
### Summary

Currently, `macOS 12` has issues with pod installation. It fails due to the `glog` dependency. This changes the runner version to `macOS 13`. I've also created a PR for using the `macOS 14` runners, however seems like the availability of those machines are low at the moment so the CI takes a long time to run.

https://github.com/callstack/react-native-builder-bob/issues/518
https://github.com/expo/expo/issues/25905

### Test plan

We need to make sure macos related runners pass on CI.

### DIFF
--- a/.github/workflows/build-templates.yml
+++ b/.github/workflows/build-templates.yml
@@ -22,8 +22,8 @@ jobs:
       fail-fast: false
       matrix:
         os:
-          - ubuntu
-          - macos
+          - ubuntu-latest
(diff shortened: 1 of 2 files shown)


atlj:
Closed this one because macOS 14 runners are better

andy380743909:
I guess the root cause of this issue is the version of Xcode not the macOS version.
For more details, see https://github.com/facebook/react-native/pull/41058

PR #525, opened 2024-02-04
feat: add a dummy swift file to the example if the library uses swift

mfazekas:
### Summary

It seems that swift debugging in libraries doesn't really work unless the app itself has a swift file. See https://github.com/callstack/react-native-builder-bob/discussions/517

There is an empty `File.swift` in example-legacy template but it's not added to the project. This change executes a ruby script to include that file in the Xcode project.

Note that it requires working `ruby` install and `xcodeproj` gem available and only run on Mac. Which is reasonable as , ruby and xcodeproj are decencies of cocoa pods.

### DIFF
--- a/packages/create-react-native-library/src/add-dummy-swift-file.rb
+++ b/packages/create-react-native-library/src/add-dummy-swift-file.rb
@@ -1,0 +1,16 @@
+gem 'xcodeproj'
+require 'xcodeproj'
+
+project_path = ARGV[0]
+target_name = ARGV[1]
+project = Xcodeproj::Project.open(project_path)
+
+project.targets.each do |target|
+  if target.name == target_name
+    swift_file = File.join(project.path, '..', 'File.swift')
+    group = project.groups.select {|i| i.name == target_name }.first
+    file = group&.new_file(swift_file)
+    target.add_file_references([file])
+  end
+end
+project.save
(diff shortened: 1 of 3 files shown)


atlj:
Hey @mfazekas thanks a lot for the PR! I was able to validate the issue on my end.
I have one question regarding the implementation, do you think it would make sense to use the [@bacons/xcode](https://github.com/EvanBacon/xcode/blob/main/README.md#api) to create the swift file instead of using ruby here? Yes we already have the necessary dependencies to run a ruby script here but most of the logic is already inside NodeJS I think directly using NodeJS instead of spawning ruby would make more sense since:

1. So far we've been using ~99% Javascript
2. Most of the contributors are familiar with Javascript
3. We have tools to statically check the Javascript code in this repo but not Ruby.
4. Non-MacOS systems would be able to create the file (This one still needs to be tested). So users can start their library in a non-darwin platform and then move to darwin. This might prevent an edge case.
5. We would like to add similar logic to modify the generated XCode project further, so this would be a great first step.

[You can find the exact same implementation done in @bacons/xcode here](https://github.com/EvanBacon/xcode/blob/main/README.md#api).
Screenshot 2024-05-24 at 21 51 06

mfazekas:
AppDelegate is now a swift file so this is pointless

PR #530, opened 2024-02-19
fix: podspec file indentation

olessavluk:
### Summary

After https://github.com/callstack/react-native-builder-bob/pull/358 identation become shifted which makes code hard to read and could lead to bugs. For example I was placing package dependencies inside `if/else` block, right after `s.dependency "React-Core"`

see 
| before | after |
|--------|--------|
| image | image |

### Test plan

Purely aesthetic changes, so nothing to break


olessavluk:
cc @ykhateeb

ykhateeb:
Thanks, @olessavluk , looks much better

ykhateeb (approved):
Thanks, @olessavluk, looks much better

satya164 (approved):
Thanks for the PR

PR #546, opened 2024-04-11
Add abiFilters to the native template

j-piasecki:
### Summary

In the native template `abiFilters` is missing from `externalNativeBuild` block. This means that the library would be built for every available ABI when building the application, even if the application itself is being built for a specific one. This is not that problematic (besides taking more time) when using React Native aar, but when building RN from source the core `.so` files will be built for that specific ABI. In case the library depends on a `.so` file, it would fail at the linking stage due to missing `.so` for the remaining ABIs.

### Test plan
See
- https://github.com/Expensify/App/pull/40102
- https://github.com/Expensify/react-native-live-markdown/pull/284
- https://github.com/margelo/react-native-quick-sqlite/pull/41

### DIFF
--- a/packages/create-react-native-library/templates/native-common/android/build.gradle
+++ b/packages/create-react-native-library/templates/native-common/android/build.gradle
@@ -18,6 +18,11 @@ buildscript {
   }
 }
 
+def reactNativeArchitectures() {
+  def value = project.getProperties().get("reactNativeArchitectures")
+  return value ? value.split(",") : ["armeabi-v7a", "x86", "x86_64", "arm64-v8a"]
+}
+
 def isNewArchitectureEnabled() {
   return rootProject.hasProperty("newArchEnabled") && rootProject.getProperty("newArchEnabled") == "true"
 }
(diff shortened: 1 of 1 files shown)


atlj (approved):

tomekzaw:
@atlj Looks like all CIs failed because of `Java heap space`, can you please re-run failed jobs?

packages/create-react-native-library/templates/native-common/android/build.gradle
```diff
@@ -18,6 +18,11 @@ buildscript {
   }
 }
 
+def reactNativeArchitectures() {
+  def value = project.getProperties().get("reactNativeArchitectures")
```
satya164, line 22: Shouldn't this be `rootProject`?

```suggestion
  def value = rootProject.getProperties().get("reactNativeArchitectures")
```

satya164 (changes requested):

satya164:
@tomekzaw seems it's failing for all of the Android builds so it doesn't seem like flaky tests but an issue with the change.

PR #551, opened 2024-04-26
feat(crnl): store metadata with new projects

atlj:
### Summary

We need a way to track the project information since we are going to introduce a library upgrade helper. With this PR, when you create a new library using `create-react-native-library`, it will automatically save;

1. Version of `create-react-native-library` used to create the project
2. The library type (native module, native view, etc.)
3. The languages used (Kotlin, objective-c, etc.)

into `create-react-native-library.json` file at the root.

### Test plan

1. Build and run `create-react-native-library`
2. Note down the options you have picked
3. Make sure the `create-react-native-library.json` file has been generated at the root of your library
4. Check the version, library type, and languages fields and make sure they are using correct data.


tsconfig.json
```diff
@@ -1,4 +1,9 @@
 {
+  "include": [
+    "packages/create-react-native-library/src",
+    "packages/create-react-native-library/package.json",
```
atlj, line 4: `import { version as crnlVersion } from '../package.json';` This line was generating errors during typecheck without including.

packages/create-react-native-library/src/index.ts
```diff
@@ -787,6 +790,41 @@ async function create(argv: yargs.Arguments<any>) {
...
+    'author-email',
+    'author-url',
+    'repo-url',
+    'example',
+    'react-native-version',
+  ];
+  const ignoredAnswers = ignoredArgs.map((argName) => questions[argName]?.name);
+  const standardArgs = ['_', '$0', 'name'];
+
+  const filteredAnswers = Object.fromEntries(
```
satya164, line 807: Maybe some name like `libraryMetadata` to make it more obvious what it is?

```suggestion
  const libraryMetadata = Object.fromEntries(
```

satya164 (changes requested):

satya164 (approved):

szymonrybczak (approved):
Works correctly! :shipit: 

```json
  "create-react-native-library": {
    "type": "module-mixed",
    "languages": "cpp",
    "version": "0.36.0"
  }
```

PR #566, opened 2024-06-14
feat: ship codegen-generated specs

atlj:
## Summary

So far with the new architecture-supported templates, we've been generating libraries that didn't ship their codegen-generated specs. This means the library's users had to build the codegen specs on their end (this was done implicitly). With this, the codegen-generated specs are generated at the build time and they are shipped with the library.

I've followed [this guide from the React Native new arch working group](https://github.com/reactwg/react-native-new-architecture/blob/main/docs/codegen.md#including-generated-code-into-libraries).

### Making sure example app builds are triggering codegen

An important problem to figure out was to make sure the codegen-generated specs were being built with each native build of the example app. To do that,`create-react-native-library` now modifies non-legacy example apps and adds:

1. A new task to `example/android/build.gradle` that's triggered before each native build.
2. A prebuild action to the  XCode build schema.
3. A `pre_install` hook to `example/ios/Podfile` that's triggered when user calls `pod install`.

These modifications make sure `yarn codegen` is called on the repo root to generate the codegen specs.

### Notes

1. There is an important problem with React Native itself right now. When `react-native codegen` is called, the generated Java code doesn't follow the `codegenConfig.android.javaPackageName` property in the `package.json`. This means the generated Java files are stored in a default location with the wrong package name. To fix it, I've added a script that moves the codegen-generated files into the correct place. You can check https://github.com/facebook/react-native/issues/45079 to see more.

### Test if Android builds trigger codegen

1. Create a non-legacy library using `create-react-native-library`
2. Install the dependencies and build the example app on Android
3. Make sure the build passes, and there are files under `android/generated`.

### Test if installing pods triggers codegen

1. Create a non-legacy library using `create-react-native-library`
2. Install the dependencies and run `pod install` in `example/ios`
3. Make sure the pods are installed and there are files under `ios/generated`.

### Test if iOS builds trigger codegen

1. Create a non-legacy library using `create-react-native-library`
2. Install the dependencies and run `pod install` in `example/ios`
3. Remove the codegen generated code from `ios/generated` since that's generated by the pod install step
4. Build the app for iOS
5. Make sure there are files in `ios/generated`.

### Test if building the library triggers codegen

1. Run `yarn prepare` to emulate the library release process
2. Run `yarn pack`
3. Extract the files from the generated `package.tgz`
4. Make sure there are files under `ios/generated`, and `android/generated` in the generated package.


packages/create-react-native-library/src/utils/patchExampleAppCodegen.ts
```diff
@@ -0,0 +1,94 @@
...
+            ActionType = "Xcode.IDEStandardExecutionActionsCore.ExecutionActionType.ShellScriptAction">
+            <ActionContent
+               title = "Invoke Codegen"
+               scriptText = "cd &quot;$WORKSPACE_PATH/../../../&quot; &amp;&amp; yarn codegen&#10;">
+            </ActionContent>
+         </ExecutionAction>
+      </PreActions>`;
+
+const PODSPEC_INVOKE_CODEGEN_SCRIPT = `
+  pre_install do |installer|
```
satya164, line 30: Can we add a comment here saying why we have it?

atlj:
move the codegen script and codegen patching into bob.

mfazekas:
> So far with the new architecture-supported templates, we've been generating libraries that didn't ship their codegen-generated specs. This means the library's users had to build the codegen specs on their end (this was done implicitly). With this, the codegen-generated specs are generated at the build time and they are shipped with the library.

I think it would be great to explain the advantages (or disadvantages) of this approach. My understand for example if you want to make android side work with old arch, the codgen generated [interfaces, delegates](https://github.com/software-mansion/react-native-screens/blob/main/android/src/paper/java/com/facebook/react/viewmanagers/RNSScreenManagerDelegate.java) can be usefull to be shipped with the project.

packages/react-native-builder-bob/src/utils/patchCodegen.ts
```diff
@@ -0,0 +1,89 @@
...
+ */
+export async function patchCodegen(projectPath: string) {
+  const packageJsonPath = path.resolve(projectPath, 'package.json');
+  const packageJson = await fs.readJson(packageJsonPath);
+
+  let codegenAndroidPath: string | undefined =
+    packageJson.codegenConfig.outputDir.android;
+  if (!codegenAndroidPath) {
+    throw new Error(
+      'You need to define codegenConfig.outputDir.android in your package.json'
```
satya164, line 20: Let's add some links to the documentation of how `codegenConfig` should look like

packages/react-native-builder-bob/src/targets/codegen.ts
```diff
@@ -0,0 +1,42 @@
...
+type Options = Input;
+
+export default async function build({ root, report }: Options) {
+  try {
+    const packageJsonPath = path.resolve(root, 'package.json');
+    if (!(await fs.pathExists(packageJsonPath))) {
+      throw new Error(
+        `Couldn't find a 'package.json' file in '${root}'. Are you in a project folder?`
+      );
+    }
```
satya164, line 16: Let's move it to the build command

packages/react-native-builder-bob/src/targets/codegen.ts
```diff
@@ -0,0 +1,42 @@
...
+
+    spawn.sync('npx', ['react-native', 'codegen'], {
+      stdio: 'inherit',
+    });
```
satya164, line 20: Let's move to async. And hide logs by default (maybe `--verbose` mode)?

packages/react-native-builder-bob/src/index.ts
```diff
@@ -471,6 +482,11 @@ yargs
     };
 
     for (const target of options.targets!) {
+      const targetArg = argv.target;
+      if (targetArg && target !== targetArg) {
+        continue;
+      }
```
satya164, line 488:
```suggestion
      if (argv.target != null) {
        buildTarget(argv.target)
      } else {
        for (const target of options.targets!) {
          buildTarget(target)
        }
      }
```

packages/create-react-native-library/templates/native-view-new/react-native.config.js
```diff
@@ -0,0 +1,12 @@
+module.exports = {
+  dependency: {
+    platforms: {
+      /**
+       * @type {import('@react-native-community/cli-types').AndroidDependencyParams}
+       */
```
satya164, line 6: Move to top to type the whole object

packages/create-react-native-library/src/utils/patchExampleAppCodegen.ts
```diff
@@ -0,0 +1,98 @@
...
+    system("cd ../../ && npx bob build --target codegen")
+  end
+`;
+
+/**
+ * Codegen isn't invoked for libraries with `includesGeneratedCode` set to `true`.
+ * This patches the example app to invoke library codegen on every app build.
+ */
+export async function patchExampleAppCodegen(
```
satya164, line 43:
```suggestion
export async function addCodegenBuildscript(
```

satya164 (changes requested):

packages/react-native-builder-bob/src/utils/patchCodegen.ts
atlj: idea: https://github.com/react-navigation/react-navigation/blob/fdbfe7608c3c9ba4c2c3c95dcc5a6501ec97a056/packages/core/src/useFocusEffect.tsx#L19

packages/react-native-builder-bob/src/targets/codegen.ts
```diff
@@ -0,0 +1,42 @@
...
+
+    patchCodegen(root);
+
+    report.success('Codegen patched successfully!');
```
satya164, line 24: Let's change the message since the message after the whole generation process and not only the patching

```suggestion
    report.success('Generated native code with codegen');
```

packages/react-native-builder-bob/src/targets/codegen.ts
```diff
@@ -0,0 +1,42 @@
...
+      } else if ('message' in e && typeof e.message === 'string') {
+        report.error(e.message);
+      } else {
+        throw e;
+      }
+    } else {
+      throw e;
+    }
+
+    throw new Error('Failed generate codegen files.');
```
satya164, line 40: I'd change the terminology a bit, as we're using code generation to generate native files - not generating code generation files themseleves

satya164 (approved):

zoontek:
@atlj @satya164 Note that, if like me you are using bob to build your Expo plugin too, it breaks as the Expo `findUpPlugin`, in charge of locating the nearest `package.json`, will now stop at the current directory (`dist/commonjs`).

- https://github.com/zoontek/react-native-bootsplash/blob/master/app.plugin.js
- https://github.com/zoontek/react-native-bootsplash/blob/master/src/expo.ts

PR #584, opened 2024-07-04
fix: devDependencies doesn't exist in rootPackageJson

iMokhles:
i faced the following issue:

Cannot set properties of undefined (setting 'react') /Users/imokhles/.nvm/versions/node/v20.13.1/lib/node_modules/create-react-native-library/lib/index.js:558
    rootPackageJson.devDependencies.react = examplePackageJson.dependencies.react;
                                          ^

TypeError: Cannot set properties of undefined (setting 'react')
    at Object.create [as handler] (/Users/imokhles/.nvm/versions/node/v20.13.1/lib/node_modules/create-react-native-library/lib/index.js:558:43)

### DIFF
--- a/packages/create-react-native-library/src/index.ts
+++ b/packages/create-react-native-library/src/index.ts
@@ -802,6 +802,10 @@ async function create(_argv: yargs.Arguments<any>) {
       path.join(folder, 'package.json')
     );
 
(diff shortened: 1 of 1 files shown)


satya164 (approved):
Thank you

PR #572, opened 2024-06-27
feat!: add ability to use `react-native-test-app` for `example/` app

szymonrybczak:
### Summary

Use [`react-native-test-app`](https://github.com/microsoft/react-native-test-app/) for `exmaple/` app. RNTA was created to help library maintainers focus on building library by providing seamless React Native versions switching, first-class support for OOT platforms and a lot more! 

### Test plan

Generated app should work in the same way as before (created library should be automatically linked).

CleanShot 2024-06-27 at 16 51 30@2x

PR #585, opened 2024-07-05
fix: fix `build:ios/android` commands in test-app example

szymonrybczak:
### Summary

In https://github.com/callstack/react-native-builder-bob/pull/572 we've introduced `test-app` as an option for `example` project, and CRNL contains two scripts `build:ios` and `build:android` that are injected when creating project, RNTA also contains two scripts named in the same way, which are then replaced - in this Pull Request in relevant scenario I added logic which joins two scripts.

### Test plan

1. Create a new project with `create-react-native-library`
2. Choose `test-app` for `example/`
3. `build:ios` and `build:android` should contain logic of creating and bundling app as well as building app with relevant options.


satya164 (approved):
Thanks

PR #594, opened 2024-07-22
fix: hide select example question when creating local library

szymonrybczak:
### Summary

A regression introduced in https://github.com/callstack/react-native-builder-bob/pull/572. 

CRNL shouldn't ask for example type in local library scenario.

### Test plan

1. Run `create-react-native-library` under a React Native project.
2. Example project shouldn't be included in the local library.


satya164 (approved):
Thank you

PR #596, opened 2024-07-23
fix: add correct script on Android

szymonrybczak:
### Summary

Fixes a regression where `build-android` contained a script for building iOS app. 

### Test plan

`build:android` should contain script for building Android app.

### DIFF
--- a/packages/create-react-native-library/src/utils/generateExampleApp.ts
+++ b/packages/create-react-native-library/src/utils/generateExampleApp.ts
@@ -163,7 +163,7 @@ export default async function generateExampleApp({
     const iosBuild = [
       'npm run mkdist',
       'react-native bundle --entry-file index.js --platform ios --dev true --bundle-output dist/main.ios.jsbundle --assets-dest dist',
-      SCRIPTS_TO_ADD['build:android'],
+      SCRIPTS_TO_ADD['build:ios'],
     ].join(' && ');
 
     Object.assign(scripts, {


packages/create-react-native-library/src/utils/generateExampleApp.ts
```diff
@@ -143,9 +143,9 @@ export default async function generateExampleApp({
   delete scripts.lint;
 
   const SCRIPTS_TO_ADD = {
-    'build:android':
+    android:
       'react-native build-android --extra-params "--no-daemon --console=plain -PreactNativeArchitectures=arm64-v8a"',
-    'build:ios': `react-native build-ios --scheme ${project.name}Example --mode Debug --extra-params "-sdk iphonesimulator CC=clang CPLUSPLUS=clang++ LD=clang LDPLUSPLUS=clang++ GCC_OPTIMIZATION_LEVEL=0 GCC_PRECOMPILE_PREFIX_HEADER=YES ASSETCATALOG_COMPILER_OPTIMIZATION=time DEBUG_INFORMATION_FORMAT=dwarf COMPILER_INDEX_STORE_ENABLE=NO"`,
+    ios: `react-native build-ios --scheme ${project.name}Example --mode Debug --extra-params "-sdk iphonesimulator CC=clang CPLUSPLUS=clang++ LD=clang LDPLUSPLUS=clang++ GCC_OPTIMIZATION_LEVEL=0 GCC_PRECOMPILE_PREFIX_HEADER=YES ASSETCATALOG_COMPILER_OPTIMIZATION=time DEBUG_INFORMATION_FORMAT=dwarf COMPILER_INDEX_STORE_ENABLE=NO"`,
   };
```
satya164, line 149: These scripts are assigned to `package.json` directly so we need to keep the original naming not to override the `android`/`ios` scripts.

szymonrybczak: It's actually assigned here: https://github.com/callstack/react-native-builder-bob/blob/0bb389badb71465ef94609fca4d37a3281136f3b/packages/create-react-native-library/src/utils/generateExampleApp.ts#L169-L172. So this change doesn't affect the final keys that are added to `package.json` 👍

satya164: @szymonrybczak also assigned here https://github.com/callstack/react-native-builder-bob/blob/0bb389badb71465ef94609fca4d37a3281136f3b/packages/create-react-native-library/src/utils/generateExampleApp.ts#L152

szymonrybczak: ah, right. fixed

satya164 (approved):
Thanks

PR #598, opened 2024-07-26
fix: always create package.json with type regardless of esm option

satya164:
This will provide more flexibilty to library authors such as specifying `type: module` in their project's `package.json` without breaking the commonjs output for the published code.

### DIFF
--- a/docs/pages/esm.md
+++ b/docs/pages/esm.md
@@ -16,12 +16,9 @@ You can verify whether ESM support is enabled by checking the configuration for 
 }
 ```
 
-The `"esm": true` option enables ESM-compatible output. Here's what it does:
+The `"esm": true` option enables ESM-compatible output by adding the `.js` extension to the import statements in the generated files.
 
-- It adds the `.js` extension to the import statements in the generated files.
-- It creates a `package.json` file in the output directory with the content: `{ "type": "module" }`
-
-In addition, it's necessary to specify `"moduleResolution": "Bundler"` in your `tsconfig.json` file:
+It's recommended to specify `"moduleResolution": "Bundler"` in your `tsconfig.json` file as well:
 
 ```json
 {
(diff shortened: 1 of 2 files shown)


atlj (approved):

zoontek:
@atlj @satya164 Note that, if like me you are using bob to build your Expo plugin too, it breaks as the Expo `findUpPlugin` function, in charge of locating the nearest `package.json`, will now stop at the current directory (`dist/commonjs`) as there's a `package.json` in it.

- https://github.com/zoontek/react-native-bootsplash/blob/master/app.plugin.js
- https://github.com/zoontek/react-native-bootsplash/blob/master/src/expo.ts

And it will break prebuild:

```
CommandError: Package "react-native-bootsplash" does not contain a valid config plugin.
Learn more: https://docs.expo.dev/guides/config-plugins/#creating-a-plugin

Cannot use import statement outside a module
```

satya164:
> @atlj @satya164 Note that, if like me you are using bob to build your Expo plugin too, it breaks as the Expo `findUpPlugin` function, in charge of locating the nearest `package.json`, will now stop at the current directory (`dist/commonjs`) as there's a `package.json` in it.
> 
> 
> 
> - https://github.com/zoontek/react-native-bootsplash/blob/master/app.plugin.js
> 
> - https://github.com/zoontek/react-native-bootsplash/blob/master/src/expo.ts
> 
> 
> 
> And it will break prebuild:
> 
> 
> 
> ```
> 
> CommandError: Package "react-native-bootsplash" does not contain a valid config plugin.
> 
> Learn more: https://docs.expo.dev/guides/config-plugins/#creating-a-plugin
> 
> 
> 
> Cannot use import statement outside a module
> 
> ```

Seems like a bug to be fixed there as this is also a pattern in other build tools like tshy

zoontek:
@satya164 How? By modifying the `findUpPlugin` fn to check for `main` `exports` properties?

satya164:
> @satya164 How? By modifying the `findUpPlugin` fn to check for `main` or `exports` properties?

Not familiar with why it looks for package.json, but if it's trying to find a package, it should check that the package.json has a name property.

PR #659, opened 2024-10-23
fix: add react-native.config.js to the published file list

johnf:
### Summary

I typed .json instead of .js in the last commit


atlj:
Thanks for this, sorry for keeping it hanging, I was trying to make sure this is absolutely needed, seems like this can solve #692. I'm going to merge and publish this

PR #668, opened 2024-10-26
feat: add `--example none`, `--replace-directory`, and `--skip-git` options

gabrielmoreira:
Hello! First, thank you for the work on Bob and create-react-native-library.

I've made a few modifications on `create-react-native-library` to support a tooling ([qdk](https://github.com/gabrielmoreira/qdk/)) setup I'm working on.

### Summary of Changes:
1. **Added `--example none` option**: Allows creating libraries without an example app.
2. **Added `--replace-directory` option**: Enables overwriting the template in an existing directory, using a similar option name as in the React Native CLI.
3. **Added `--skip-git` option**: Prevents any git commands from running during template creation, also following a naming convention similar to the React Native CLI.

### Context:
In my tool ([qdk](https://github.com/gabrielmoreira/qdk/blob/main/templates/monorepo/.qdk/components/ReactNativeLibProject.ts#L238)), I have templates for different project structures, and one of these templates uses `create-react-native-library` to set up a React Native library project. The goal is to generate only the library code, without an example app, without interfering with the project git, and with the flexibility to overwrite an existing directory if it already contains some files.

Thanks for considering these additions!

*P.S.*: By the way, the options added here follow a similar naming convention to `npx @react-native-community/cli init` options, for consistency.

I also considered contributing an optional parameter to fix the version of `npx @react-native-community/cli@<<VERSION>> init`, but for my case, not creating the example app was more beneficial. The vanilla example was giving me errors during generation, even though it was working a few days ago. Fixing the version of create-react-library wasn’t enough.

Additionally, I thought about adding an option to specify the package manager instead of trying to detect it. However, this would require more testing with various package managers, so I decided to hold off on that for now.

### DIFF
--- a/packages/create-react-native-library/src/index.ts
+++ b/packages/create-react-native-library/src/index.ts
@@ -160,6 +160,11 @@ const EXAMPLE_CHOICES = [
     value: 'expo',
     description: 'managed expo project with web support',
   },
+  {
+    title: 'None',
+    value: 'none',
+    description: 'no example app will be created',
+  },
 ] as const;
 
 const NEWARCH_DESCRIPTION = 'requires new arch (experimental)';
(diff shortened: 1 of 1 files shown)


gabrielmoreira:
Hi folks! This PR is a small, focused change with three simple commits. If someone could take a quick look, I'd really appreciate it!

satya164 (changes requested):
Thanks for the PR. I'd like to make following changes:

- To match the convention for booleans, instead of `skip-git`, add a `git` boolean option, and default it to true. To disable git, user would be able to pass `--no-git`
- I'd remove `replace-directory` option. I'd like to avoid additional option for this since you can just delete the folder beforehand in your code.

satya164:
Closing as it's stale

PR #679, opened 2024-11-08
feat: don't add a XCode prebuild action to invoke codegen anymore

atlj:
### Summary

We used to add an xcode prebuild script to invoke codegen each time the user built their example application. However, with the latest setup, the action just causes the build to fail immediately. This PR removes that script. With this removal, the user has to install pods each time they change their codegen specs.

### Test plan

1. Create a new library using `crnl` and make sure it supports the new architecture
2. Install the pods and try building the example app using xcode
3. Make sure the build doesn't fail immediately.


atlj:
> What's the reason for this? Is it not possible to fix it?

It cannot find `npx`, which is needed to invoke bob's codegen build command. I also tried to use `npm exec` but it couldn't find `npm` too. I also tried to use React Native's `WITH_ENVIRONMENT.sh`, and it's able to find node but `npx` and `npm` are still missing. Also this step was causing a lot of issues for users. So I think it's the best to remove it.

> At the very least, is there any way to ensure that user knows that they need to run pod install first?

The user actually always needed to call `pod install` if they changed the codegen specs, even with the previous setup where we didn't generate codegen for them. Codegen is invoked for iOS during `pod install`. So this won't affect the existing workflow.

> Is it possible to fail the build in case of outdated generated code?

The build is going to fail if the user tries to build their example app with outdated codegen scaffold code.

satya164:
> It cannot find npx, which is needed to invoke bob's codegen build command. I also tried to use npm exec but it couldn't find npm too.

Why do we need either of those? Can't we just check `node_modules/.bin`?

atlj:
> Can't we just check node_modules/.bin?

I just checked and it's a great idea, however, I had lots of difficulties just trying to get it working on my machine:

1. I first tried to call bob directly from `.bin` without doing anything else and I got an error saying `node not found`. Then I tried to use React Native's script to inject the `node` binary into env but I kept getting the same error.

And also, looks like with XCode 14, we need to patch the file further to get the React Native vars:

```diff
diff --git a/example/ios/BuraktestExample.xcodeproj/xcshareddata/xcschemes/BuraktestExample.xcscheme b/example/ios/BuraktestExample.xcodeproj/xcshareddata/xcschemes/BuraktestExample.xcscheme
index 02d793e..b54af5f 100644
--- a/example/ios/BuraktestExample.xcodeproj/xcshareddata/xcschemes/BuraktestExample.xcscheme
+++ b/example/ios/BuraktestExample.xcodeproj/xcshareddata/xcschemes/BuraktestExample.xcscheme
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <Scheme
    LastUpgradeVersion = "1210"
-   version = "1.3">
+   version = "1.7">
    <BuildAction
       parallelizeBuildables = "YES"
       buildImplicitDependencies = "YES">
@@ -11,6 +11,15 @@
             <ActionContent
                title = "Invoke Codegen"
                scriptText = "WITH_ENVIRONMENT=&quot;$REACT_NATIVE_PATH/scripts/xcode/with-environment.sh&quot;&#10;&#10;cd &quot;$WORKSPACE_PATH/../../../&quot; &amp;&amp; /bin/sh -c &quot;$WITH_ENVIRONMENT &apos;./node_modules/.bin/bob build --target codegen&apos;&quot;&#10;">
+               <EnvironmentBuildable>
+                  <BuildableReference
+                     BuildableIdentifier = "primary"
+                     BlueprintIdentifier = "13B07F861A680F5B00A75B9A"
+                     BuildableName = "BuraktestExample.app"
+                     BlueprintName = "BuraktestExample"
+                     ReferencedContainer = "container:BuraktestExample.xcodeproj">
+                  </BuildableReference>
+               </EnvironmentBuildable>
             </ActionContent>
          </ExecutionAction>
       </PreActions>
```

I think we can get this working with additional work, but this might be broken in the future as it relies on the XCode version, and the React Native version. So I'm not sure if it's worth the effort to update and maintain this script.

satya164:
How does React Native find node to build the JS bundle, launch the dev server etc. when building with XCode? What's different in our implementation?

atlj:
React Native uses [a shell script](https://github.com/facebook/react-native/blob/main/packages/react-native/scripts/react-native-xcode.sh) call various scripts. they inject the `$NODE_BINARY` from [another script](https://github.com/facebook/react-native/blob/main/packages/react-native/scripts/xcode/with-environment.sh) which uses an artifact from the `pod install` step.

satya164:
Can't we use this `NODE_BINARY` as well?

atlj:
We tried to automate the process using https://github.com/CocoaPods/Xcodeproj, however, it's taking too much time to get everything right for all of the users. We're sunsetting this feature for now and we can return to this in the future.

Issue #676, opened 2024-11-06
Yarn Prepare failure


bmahajanZ:
### Description

As mentioned here - https://github.com/facebook/react-native/issues/47353#issuecomment-2455219536, opening a new bug

### Packages

- [ ] create-react-native-library
- [x] react-native-builder-bob

### Selected options

npx create-react-native-library@latest <Name of Your Library>

### Link to repro

https://github.com/mbhaskar98/test-library

### Environment

```bash
❯ npx react-native info

⚠️ react-native depends on @react-native-community/cli for cli commands. To fix update your package.json to include:


  "devDependencies": {
    "@react-native-community/cli": "latest",
  }




/test-library  on main +52 ?4 ──────────────────── at 15:26:23 
❯ npx @react-native-community/cli@latest info                             
info Fetching system and libraries information...
System:
  OS: macOS 14.7
  CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
  Memory: 83.04 MB / 16.00 GB
  Shell:
    version: "5.9"
    path: /bin/zsh
Binaries:
  Node:
    version: 20.11.1
    path: ~/.nvm/versions/node/v20.11.1/bin/node
  Yarn:
    version: 3.6.1
    path: ~/.nvm/versions/node/v20.11.1/bin/yarn
  npm:
    version: 10.2.4
    path: ~/.nvm/versions/node/v20.11.1/bin/npm
  Watchman:
    version: 2024.06.10.00
    path: /usr/local/bin/watchman
Managers:
  CocoaPods:
    version: 1.15.2
    path: /Users/bhaskarmahajan/.rbenv/shims/pod
SDKs:
  iOS SDK:
    Platforms:
      - DriverKit 23.5
      - iOS 17.5
      - macOS 14.5
      - tvOS 17.5
      - visionOS 1.2
      - watchOS 10.5
  Android SDK: Not Found
IDEs:
  Android Studio: 2024.2 AI-242.23339.11.2421.12483815
  Xcode:
    version: 15.4/15F31d
    path: /usr/bin/xcodebuild
Languages:
  Java:
    version: 21.0.2
    path: /usr/bin/javac
  Ruby:
    version: 3.3.3
    path: /Users/bhaskarmahajan/.rbenv/shims/ruby
npmPackages:
  "@react-native-community/cli": Not Found
  react:
    installed: 18.3.1
    wanted: 18.3.1
  react-native:
    installed: 0.76.1
[...]

PR #684, opened 2024-11-15
fix: every created example app is testing app

atlj:
### Summary

We aren't properly injecting the single-choice questions into our answers yet, and as a result the default example app was testing app. Addressed that problem

### Test plan

1. Create a new library
2. Make sure the example app is a vanilla testing app


atlj:
Fixes #676

PR #685, opened 2024-11-15
feat: use node to call codegen

atlj:
### Summary

We were using `npx` to call the React Native Community CLI
for codegen. However, not every environment has `npx` available. This
adds custom logic to invoke the RNCCLI only using the node binary.

This also emits a better error message when the user doesn't have `@react-native-community/cli` installed on their library project.

### Test plan
1. Create a new architecture supported library
3. Call `yarn bob build --target codegen`
3. Make sure the command passes without any problems


atlj:
This will emit a much better error message in cases like #676

PR #688, opened 2024-11-16
refactor(crnl): split the index file

atlj:
### Summary

The previous `index.ts` file of `create-react-native-library` was too long and complicated to maintain. This PR splits the file a bit

### Test plan

1. We have to make sure that the behavior hasn't changed at all. CI pipeline is the best way to check this.


packages/create-react-native-library/src/exampleApp/dependencies.ts
```diff
@@ -0,0 +1,47 @@
+import assert from 'node:assert';
+import path from 'path';
+import fs from 'fs-extra';
+import type { ExampleApp } from '../input';
+
+export async function getDependencyVersionsFromExample(
```
atlj, line 6:
```suggestion
export async function getDependencyVersionsFromExampleApp(
```

packages/create-react-native-library/src/exampleApp/dependencies.ts
```diff
@@ -0,0 +1,47 @@
...
+  folder: string,
+  exampleAppType: ExampleApp
+) {
+  // Set `react` and `react-native` versions of root `package.json` from example `package.json`
```
atlj, line 10:
```suggestion
```

PR #689, opened 2024-11-16
fix: spawn CLI directly without using Node

gronxb:
### Summary
image
The default is set to `yarn`, but an error occurs when using `pnpm` to run `bob build`.

The error arises because the CLI is spawned via Node. Since the CLI in `.bin` is designed to run with Node by default, it should spawn correctly even without explicitly prepending node.

### Test plan

> The changes work correctly even when tested with Yarn.

1. Create a repository using the command:
```sh
> npx create-react-native-library@latest rn-lib
```
3. Remove the `packageManager` field from the `package.json`.
4. Run `pnpm install`
* as-is
5. An error occurs.
* to-be
5. Success.

### DIFF
--- a/packages/react-native-builder-bob/src/utils/runRNCCli.ts
+++ b/packages/react-native-builder-bob/src/utils/runRNCCli.ts
@@ -27,7 +27,7 @@ export async function runRNCCli(
     rncCliBinaryName
   );
 
-  return await spawn(RNC_CLI_BINARY_PATH, [...args], options);
+  return await spawn(RNC_CLI_BINARY_PATH, args, options);
 }
 
 async function getCliBinaryName(): Promise<string> {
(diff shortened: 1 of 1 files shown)


atlj:
Hey, thanks for the issue, for looks like `pnpm` treats the binaries differently than `yarn` and this is why we are having this problems. Here's the bob binary generated with `pnpm`:

```sh
#!/bin/sh
basedir=$(dirname "$(echo "$0" | sed -e 's,\\,/,g')")

case `uname` in
    *CYGWIN*) basedir=`cygpath -w "$basedir"`;;
esac

if [ -z "$NODE_PATH" ]; then
  export NODE_PATH="<MYPATHS>"
else
  export NODE_PATH="<MYPATHS>"
fi
if [ -x "$basedir/node" ]; then
  exec "$basedir/node"  "$basedir/../@react-native-community/cli/build/bin.js" "$@"
else
  exec node  "$basedir/../@react-native-community/cli/build/bin.js" "$@"
fi
```

and the same binary generated with `yarn`:

```js
#!/usr/bin/env node

// eslint-disable-next-line import/no-commonjs
require('../lib/index');
```

atlj (approved):

Issue #678, opened 2024-11-08
TypeScript builds ignore tsconfig compilerOptions.jsx. Output is always 'react-jsx' 


marklawlor:
### Description

Libraries built with `react-native-builder-bob` do not respect the `tsconfig.config#compilerOptions.jsx` setting, and force the output to be `react-jsx`. This is a problem for React Native libraries which normally publish as `jsx: 'react-native' or 'preserve'`.

Libraries such as `@emotion/native` and `NativeWind` both rely on setting a custom `jsxImportSource`, which only works when a library has been published with persevered JSX. This means libraries built with `react-native-builder-bob` will stop working with these libraries.

The cause of this issue is the babel-preset, which should conditionally turn off `'@babel/plugin-transform-react-jsx'` / `'@babel/plugin-transform-react-jsx-development'` when `jsx` is set to `react-native` or `preserve`.

### Packages

- [ ] create-react-native-library
- [x] react-native-builder-bob

### Selected options

-

### Link to repro

https://github.com/marklawlor/react-native-builder-bob-jsx

### Environment

-

atlj:
Fixed by #695

marklawlor:
Please reopen the issue, the `tsconfig#jsx` option is different to setting the runtime to `classic` or `automated`

Please see https://github.com/software-mansion/react-native-reanimated/issues/6665#issuecomment-2492023070 for more context

PR #695, opened 2024-11-21
feat: allow to specify JSX Runtime for @babel/preset-react`

tjzel:
### Summary

In https://github.com/callstack/react-native-builder-bob/commit/0595213e4814c9bc34c270ef89bf31f8e0bd9bed the JSX runtime was changed to `automatic` without the option for the user to opt back in for `classic`.

The `classic` runtime is needed for NativeWind library, which is dependent on Reanimated.

Fixes
- https://github.com/software-mansion/react-native-reanimated/issues/6665
- #678 

### Test plan

🚀


satya164 (approved):
Thanks for the PR

PR #697, opened 2024-11-21
fix: use proper codegen types

atlj:
### Summary

We were using `type: "all"` in codegenConfig for all types of libraries. @cipolleschi recommended us to use proper types instead [in this thread](https://github.com/facebook/react-native/issues/46208#issuecomment-2491424662). This updates the templates to use the proper types

#### Modules

1. Create a new arch supported module library
2. Go to the `package.json` and make sure `codegenConfig.type` is `"modules"`
3. Make sure the library builds and runs fine with the new architecture

#### Components a.k.a views

1. Create a new arch supported view/component
2. Go to the `package.json` and make sure `codegenConfig.type` is `"components"`
3. 3. Make sure the library builds and runs fine with the new architecture

### DIFF
--- a/packages/create-react-native-library/templates/common/$package.json
+++ b/packages/create-react-native-library/templates/common/$package.json
@@ -198,7 +198,7 @@
   },
   "codegenConfig": {
     "name": "RN<%- project.name -%><%- project.view ? 'View': '' -%>Spec",
-    "type": "all",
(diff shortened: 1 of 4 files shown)


cipolleschi (approved):
looks good. `all` is still a valid type, but should be used when a libraries exposes at least a module and at least a component.

But this is an advanced use case that perhaps users can apply themselves while developing, there is no need to control it in the template probably.

PR #698, opened 2024-11-22
fix(cli): update stdio options to capture error output

gronxb:
### Summary

If an error occurs, it is not printed in the CLI.

* as-is:
image

* to-be:
image

### Test plan
1. `react-native-config.js` file: 
```js
test // That code will cause an error.
```
2. yarn prepare


satya164 (approved):
Thanks for the PR!

PR #707, opened 2024-12-03
Fix: Ensure 'invokeLibraryCodegen' task executes 'npx' correctly on all OS

ddiazFG:
@IMPORTANT

feel free to close it if it only fails because a local configuration is incorrect.

### Summary

Fixed the error "Execution failed for task ':app:invokeLibraryCodegen'" when running yarn android in the Fabric View Library. The invokeLibraryCodegen task now correctly executes npx on all platforms by checking the OS and using cmd or sh accordingly.

### Test plan

Before the fix: Run `yarn android` on example forlder and observe the build failing with the mentioned error.
Apply the fix: Update the build script with the provided solution.
After the fix: Run yarn android again and confirm that the build succeeds without errors.


satya164 (approved):
Thanks for the PR

PR #712, opened 2024-12-03
refactor: drop backward compatible templates

satya164:
As of React Native 0.76, the new architecture is now default. So this moves the templates with new architecture up. New libraries should start with new arch template.

For legacy usage, we still have the legacy template for now. Old architecture libraries still work in new architecture via compat layer. So the backward compatible templates are unnecessary. They are also difficult to maintain and are currently broken, so I'm removing them.


maitrungduc1410:
hi @satya164 just want to confirm, library generated using `create-react-native-library@latest` should work in app with newArc and legacy right?

satya164:
@maitrungduc1410 the latest templates only support the new architecture. you will need to use an older version to generate project for old architecture (0.49.8 for legacy templates and 0.44.3 for backward compatible templates).

react native docs also say you can support old architecture on Android by using `includesGeneratedCode` but I haven't tested that. references:

https://reactnative.dev/docs/the-new-architecture/codegen-cli#enabling-includesgeneratedcode
https://callstack.github.io/react-native-builder-bob/build#codegen

maitrungduc1410:
Got it, thanks

Issue #713, opened 2024-12-03
Codegen config issue


riteshshukla04:
### Description

I am facing issue while using codegen . The build starts failing 

1) When I remove outputDir from codegen config android build starts breaking. (which I think is optional)
2) If I turn the includesGeneratedCode to false cmake is no longer able to link the specs generated 'react_codegen_RNCodegenExampleSpec'

If I keep the outputDir  and includesGeneratedCode I am getting the below error on iOS
![Image](https://github.com/user-attachments/assets/9fa13bf3-50a9-4b00-8554-7f1e65d587ee)

### Packages

- [x] create-react-native-library
- [ ] react-native-builder-bob

### Selected options

✔ What type of library do you want to develop? › Turbo module
✔ Which languages do you want to use? › C++ for Android & iOS

### Link to repro

https://github.com/riteshshukla04/rn-codegen-repro

### Environment

System:
  OS: macOS 15.1
  CPU: (8) arm64 Apple M2
  Memory: 194.16 MB / 16.00 GB
  Shell:
    version: "5.9"
    path: /bin/zsh
Binaries:
  Node:
    version: 22.7.0
    path: ~/.nvm/versions/node/v22.7.0/bin/node
  Yarn:
    version: 3.6.1
    path: ~/.nvm/versions/node/v22.7.0/bin/yarn
  npm:
    version: 10.8.2
    path: ~/.nvm/versions/node/v22.7.0/bin/npm
  Watchman: Not Found
Managers:
  CocoaPods:
    version: 1.15.2
    path: /Users/riteshshukla2/.rvm/gems/ruby-2.7.5/bin/pod
SDKs:
  iOS SDK:
    Platforms:
      - DriverKit 24.1
      - iOS 18.1
      - macOS 15.1
      - tvOS 18.1
      - visionOS 2.1
      - watchOS 11.1
  Android SDK: Not Found
IDEs:
  Android Studio: 2024.1 AI-241.18034.62.2412.12266719
  Xcode:
    version: 16.1/16B40
    path: /usr/bin/xcodebuild
Languages:
  Java:
    version: 17.0.12
    path: /usr/bin/javac
  Ruby:
    version: 2.7.5
    path: /Users/riteshshukla2/.rvm/rubies/ruby-2.7.5/bin/ruby
npmPackages:
  "@react-native-community/cli": Not Found
  react: Not Found
  react-native: Not Found
  react-native-macos: Not Found
npmGlobalPackages:
[...]

PR #716, opened 2024-12-04
fix: add optional Chaining in Android output dir

riteshshukla04:
This will prevent undefined from being thrown as error and actual error will be thrown.

https://github.com/callstack/react-native-builder-bob/issues/713


satya164 (approved):
Thanks for the PR

PR #721, opened 2024-12-06
feat(crnl): support nitro modules

atlj:
### Summary

Adds [Nitro Modules](https://nitro.margelo.com/) support to `create-react-native-library`.
Screenshot 2024-12-06 at 20 05 42

### Test plan

Prior to all cases:
1. Clone this branch
2. Call `yarn` and `yarn prepare`
3. Call `bin/create-react-native-library`

#### Example app

1. Create a new nitro module
4. Go to the module and install node dependencies
5. Call `yarn nitrogen` to create nitro bindings
6. Build the Android app and make sure it works properly
7. Install the pods and build the iOS app and make sure it works properly

#### Published lib

1. Publish the library from the previous case using `https://github.com/wclr/yalc` (`yalc publish`)
2. Create a new React Native app
3. Call `yalc add your-lib`
4. Make sure the React Native app builds and works fine with your lib.


mrousavy:
letsgoooooo!!!

packages/create-react-native-library/src/inform.ts
```diff
@@ -96,6 +96,16 @@ export async function printNextSteps(
         )
         .join('\n')}
 
+      ${
+        config.project.nitro
+          ? kleur.blue(
+              `Read the ${kleur.bold(
+                'Nitro Modules'
+              )} docs at ${kleur.underline('https://nitro.margelo.com/docs')}
```
mrousavy, line 104:
```suggestion
              )} docs at ${kleur.underline('https://nitro.margelo.com')}
```

packages/create-react-native-library/templates/common/$package.json
```diff
@@ -85,10 +91,16 @@
     "eslint-config-prettier": "^9.0.0",
     "eslint-plugin-prettier": "^5.0.1",
     "jest": "^29.7.0",
+<% if (project.nitro) { -%>
+    "nitro-codegen": "^<%- versions.nitroCodegen %>",
```
mrousavy, line 95: ah here we have it - ok good!

packages/create-react-native-library/templates/common/CONTRIBUTING.md
```diff
@@ -22,6 +22,9 @@ yarn
 The [example app](/example/) demonstrates usage of the library. You need to run it to test any changes you make.
 
 It is configured to use the local version of the library, so any changes you make to the library's source code will be reflected in the example app. Changes to the library's JavaScript code will be reflected in the example app without a rebuild, but native code changes will require a rebuild of the example app.
+<% if (project.nitro) { -%>
+This project uses [Nitro Modules](https://nitro.margelo.com/). If you're not familiar with how Nitro works, make sure to check the [Nitro Modules Docs](https://nitro.margelo.com/docs).
```
mrousavy, line 26:
```suggestion
This project uses [Nitro Modules](https://nitro.margelo.com/). If you're not familiar with how Nitro works, make sure to check the [Nitro Modules Docs](https://nitro.margelo.com/).
```

packages/create-react-native-library/templates/native-common/android/build.gradle
```diff
@@ -69,28 +76,62 @@ android {
...
         abiFilters (*reactNativeArchitectures())
+
+        buildTypes {
+          debug {
+            cppFlags "-O1 -g"
+          }
+          release {
+            cppFlags "-O2"
+          }
+        }
```
mrousavy, line 94: It doesnt _have_ to be configured like this. `-DANDROID_STL=c++_shared` is something I recommend for EVERY C++ lib, even Turbo/Expo modules, but the rest is purely up to you guys. Leaving it like this is good imo, but it's technically slower in debug because of -O1

packages/create-react-native-library/templates/nitro-module/ios/{%- project.name %}.swift
```diff
@@ -0,0 +1,10 @@
+class <%- project.name -%>: Hybrid<%- project.name -%>Spec {
+    public var hybridContext = margelo.nitro.HybridContext()
+    public var memorySize: Int {
+        getSizeOf(self)
+    }
```
mrousavy, line 5: I should really work on getting this lifted into the spec instead of having it at user class level. Maybe Nitro 19

packages/create-react-native-library/templates/native-common/android/build.gradle
atlj: @mrousavy it's going to run slower, not build slower right? Besides that I think this is a great default to have for every cpp lib, that's why I kept it this way.

packages/create-react-native-library/src/exampleApp/generateExampleApp.ts
```diff
@@ -188,12 +178,19 @@ export default async function generateExampleApp({
   });
 
   const PACKAGES_TO_ADD_DEV = {
-    'react-native-builder-bob': `^${bobVersion}`,
+    'react-native-builder-bob': `^${config.versions.bob}`,
   };
 
+  if (config.project.nitro) {
+    const packagesToAddNitro = {
+      'react-native-nitro-modules': `^${config.versions.nitroModules}`,
```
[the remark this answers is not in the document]
atlj: this is purely the example app stuff so it's not needed here

packages/create-react-native-library/templates/native-common/android/build.gradle
mrousavy: It's gonna build faster, and run slower in debug (`-O1` is less aggressive optimizations than `-O2`). :)

This also might include some more debug symbols and strips out less unused vars, might help for debugging.

.github/workflows/build-templates.yml
```diff
@@ -47,6 +48,10 @@ jobs:
             language: cpp
           - type: legacy-module
             language: cpp
+          - type: nitro-module
+            language: kotlin-objc
```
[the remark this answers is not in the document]
mrousavy: got it thanks!

packages/create-react-native-library/templates/native-common/android/build.gradle
atlj: I think that would be perfectly reasonable in this case, also builds running faster on debug would be the cherry on the top

packages/create-react-native-library/templates/native-common/android/build.gradle
```diff
@@ -69,28 +76,62 @@ android {
...
+            "**/libhermes.so",
+            "**/libhermes-executor-debug.so",
+            "**/libhermes_executor.so",
+            "**/libreactnative.so",
+            "**/libreactnativejni.so",
+            "**/libturbomodulejsijni.so",
+            "**/libreact_nativemodule_core.so",
+            "**/libjscexecutor.so"
+    ]
+  }
```
rodgomesc, line 126: this still need those days?, I created a rn turbo modules recently then migrated to nitro later, and i didn't got any symbol conflict as we used to have in the past

atlj: I've got these from @mrousavy's nitro template, @mrousavy is it safe to remove these? Also maybe we can make this to default for everyone?

.github/workflows/build-templates.yml
```diff
@@ -207,6 +212,11 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-gradle-
 
+      - name: Run nitro codegen
+        if: matrix.type == 'nitro-module'
+        run: |
+          yarn nitrogen
```
satya164, line 218: We need to run this manually?

Also contributing guide needs to include the workflow in case there are any new steps - like running this (and when is it needed)

packages/create-react-native-library/src/exampleApp/generateExampleApp.ts
```diff
@@ -231,15 +228,16 @@ export default async function generateExampleApp({
...
     app.expo.ios = app.expo.ios || {};
-    app.expo.ios.bundleIdentifier = `${project.package}.example`;
+    app.expo.ios.bundleIdentifier = `${config.project.package}.example`;
 
     await fs.writeJSON(path.join(directory, 'app.json'), app, {
       spaces: 2,
     });
   }
 
+  // Sort the deps by name
```
satya164, line 240: Isn't it evident by the name of `sortObjectKeys` function?

For comment it'd be better to say why 

```suggestion
  // Sort the deps by name to match behavior of package managers
  // This way the package.json doesn't get updated when installing deps
```

packages/create-react-native-library/src/index.ts
```diff
@@ -45,6 +51,12 @@ async function create(_argv: yargs.Arguments<Args>) {
 
   // Prefetch bob version in background while asking questions
   const resolveBobVersion = resolveBobVersionWithFallback(FALLBACK_BOB_VERSION);
+  const resolveNitroModulesVersion = resolveNitroModulesVersionWithFallback(
+    FALLBACK_NITRO_MODULES_VERSION
+  );
+  const resolveNitroCodegenVersion = resolveNitroCodegenVersionWithFallback(
+    FALLBACK_NITRO_CODEGEN_VERSION
+  );
```
satya164, line 59: Seems unnecessary to create separate function for each thing and duplicating the logic. Instead make one function `resolveNpmPackageVersionWithFallback` and use as `resolveNpmPackageVersionWithFallback(packageName, fallbackVersion)`

packages/create-react-native-library/templates/common/CONTRIBUTING.md
```diff
@@ -22,6 +22,9 @@ yarn
 The [example app](/example/) demonstrates usage of the library. You need to run it to test any changes you make.
 
 It is configured to use the local version of the library, so any changes you make to the library's source code will be reflected in the example app. Changes to the library's JavaScript code will be reflected in the example app without a rebuild, but native code changes will require a rebuild of the example app.
+<% if (project.nitro) { -%>
```
satya164, line 25: Does this add newline between the paragraphs? 🤔

packages/create-react-native-library/src/inform.ts
```diff
@@ -96,6 +96,16 @@ export async function printNextSteps(
...
+              `Read the ${kleur.bold(
+                'Nitro Modules'
+              )} docs at ${kleur.underline('https://nitro.margelo.com')}
+              `
+            )
+          : ''
+      }
       ${kleur.yellow(
         `See ${kleur.bold('CONTRIBUTING.md')} for more details. Good luck!`
       )}
```
satya164, line 111: Imo it's a bit awkward:

```
Read [..] at [...]
See [..] for more details
```

Reads like the the see for more details is talking about the link above. I'd move the nitro link to below, and add a new line between them. Or remove the link altogether since it's in CONTRIBUTING.md anyway - and people don't read that. So not sure if 2 calls to action is better.

packages/create-react-native-library/src/input.ts
```diff
@@ -84,6 +85,11 @@ const TYPE_CHOICES: {
   value: ProjectType;
   description: string;
 }[] = [
+  {
+    title: 'Nitro module',
+    value: 'nitro-module',
+    description: 'nitro for faster modules',
```
satya164, line 91: The description should say what it is for. What can I use nitro modules for? For wrapping native APIs? Writing native views? Both?

packages/create-react-native-library/src/template.ts
```diff
@@ -137,10 +142,12 @@ export function generateTemplateConfiguration({
               .replace(/[^a-z0-9](\w)/g, (_, $1) => $1.toUpperCase())
               .slice(1)}`,
       package: pack,
+      package_array: JSON.stringify(pack.split('.')),
       package_dir: pack.replace(/\./g, '/'),
       package_cpp: pack.replace(/\./g, '_'),
       identifier: slug.replace(/[^a-z0-9]+/g, '-').replace(/^-/, ''),
       native: languages !== 'js',
+      nitro: type.startsWith('nitro'),
```
satya164, line 150: Keep in mind that string matching means refactors won't be caught by TS. Prefer explicitly checking the string with equality operator.

packages/create-react-native-library/src/template.ts
```diff
@@ -137,10 +142,12 @@ export function generateTemplateConfiguration({
               .replace(/[^a-z0-9](\w)/g, (_, $1) => $1.toUpperCase())
               .slice(1)}`,
       package: pack,
+      package_array: JSON.stringify(pack.split('.')),
```
satya164, line 145: Since this is a single-use variable and is derived from another variable, prefer writing the split logic in the usage site directly instead of new config

packages/create-react-native-library/templates/common/$package.json
```diff
@@ -100,7 +112,12 @@
   },
   "peerDependencies": {
     "react": "*",
+<% if (project.nitro) { -%>
+    "react-native": "*",
+    "react-native-nitro-modules": "*"
```
satya164, line 117: Will the library stay compatible with multiple versions of `react-native-nitro-modules`?

packages/create-react-native-library/templates/native-common/android/build.gradle
```diff
@@ -25,6 +25,9 @@ def isNewArchitectureEnabled() {
 
 apply plugin: "com.android.library"
 apply plugin: "kotlin-android"
+<% if (project.nitro) { -%>
+apply from: '../nitrogen/generated/android/<%- project.package_cpp -%>+autolinking.gradle'
```
satya164, line 29: Is this file generated by `nitrogen`? Then it also needs to be in the build script so that it's run every time before publishing the package.

Also shouldn't this be in `.gitignore`?

packages/create-react-native-library/templates/native-common/android/build.gradle
satya164, line 94: If these are only needed for nitro make sure to wrap it a condition

packages/create-react-native-library/templates/native-common/android/build.gradle
satya164, line 126: If these are needed only for nitro wrap them in a condition

packages/create-react-native-library/templates/native-common/android/build.gradle
```diff
@@ -136,9 +177,13 @@ dependencies {
   //noinspection GradleDynamicVersion
   implementation "com.facebook.react:react-native:+"
   implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
+<% if (project.nitro) { -%>
+  // Add a dependency on NitroModules
```
satya164, line 181: Redundant comment

packages/create-react-native-library/templates/nitro-module/src/$index.tsx
```diff
@@ -0,0 +1,9 @@
+import { NitroModules } from 'react-native-nitro-modules';
```
satya164, line 1: What's up with the `$` prefix in the filename?

packages/create-react-native-library/templates/nitro-module/android/src/main/cpp/$cpp-adapter.cpp
```diff
@@ -0,0 +1,6 @@
+#include <jni.h>
```
satya164, line 1: What's up with the `$` prefix in the filename?

packages/create-react-native-library/templates/nitro-module/android/$CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
+project(<%- project.package_cpp -%>)
```
satya164, line 1: What's up with the `$` prefix in the filename?

packages/create-react-native-library/templates/nitro-module/$nitro.json
```diff
@@ -0,0 +1,17 @@
+{
```
satya164, line 1: What's up with the `$` prefix in the filename?

packages/create-react-native-library/templates/native-common/android/build.gradle
```diff
@@ -49,7 +52,11 @@ def supportsNamespace() {
 
 android {
   if (supportsNamespace()) {
+<% if (project.nitro) { -%>
+    namespace "com.margelo.nitro.<%- project.package -%>"
```
satya164, line 56: isn't this the library's namespace? then it shouldn't have margelo.com since that's not the library's organization.

packages/create-react-native-library/src/index.ts
```diff
@@ -73,9 +85,15 @@ async function create(_argv: yargs.Arguments<Args>) {
   assertUserInput(questions, answers);
 
   const bobVersion = await resolveBobVersion();
+  const nitroModulesVersion = await resolveNitroModulesVersion();
+  const nitroCodegenVersion = await resolveNitroCodegenVersion();
```
satya164, line 89: if user didn't select nitro then we shouldn't wait for it. when not in interactive mode it can add up to 2 second delay to the CLI

packages/create-react-native-library/templates/common/$package.json
```diff
@@ -72,7 +78,7 @@
   "devDependencies": {
     "@commitlint/config-conventional": "^17.0.2",
     "@evilmartians/lefthook": "^1.5.0",
-<% if (example === 'vanilla') { -%>
+<% if (example === 'vanilla' && !project.nitro) { -%>
```
satya164, line 81: It'll be more future-proof to add configs based on what type of project it is rather than what type of project it is not. That way adding a new type won't require adding new condition to each call site.

I think we should just add a key for the official module type and use it for these checks (here and other places) instead of `!project.nitro`.

satya164 (changes requested):

packages/create-react-native-library/templates/common/README.md
```diff
@@ -4,9 +4,15 @@
 
 ## Installation
 
+<% if (project.nitro) { -%>
+```sh
+npm install <%- project.slug %> react-native-nitro-modules
```
satya164, line 9: Probably should also add a very short description and link to nitro website below this so people know what is the `react-native-nitro-modules` package that they are installing.

packages/create-react-native-library/templates/native-common/android/build.gradle
mrousavy: I don't know, lol.

packages/create-react-native-library/templates/common/CONTRIBUTING.md
atlj: added it

packages/create-react-native-library/templates/nitro-module/$nitro.json
atlj: removed them 👍

packages/create-react-native-library/templates/common/README.md
atlj: I added:
> `react-native-nitro-modules` is required as this library relies on [Nitro Modules](https://nitro.margelo.com/).

packages/create-react-native-library/templates/native-common/android/build.gradle
atlj: These are nice to have for every type of C++ library so let's definitely keep them. An overall improvement.

packages/create-react-native-library/templates/common/$package.json
atlj: I've introduced `project.moduleConfig` (which replaces `project.module`) as an union type of `native-modules`, `turbo-modules`, `nitro-modules`. This is the native modules api that we use.

packages/create-react-native-library/templates/common/$package.json
atlj: I guess this would be a question for @mrousavy. Are you folks planning to introduce breaking changes in the future? Will it always use zero based versioning? Should we use some version constraints here?

atlj: But, I think we shouldn't worry about this too much tho, people don't care about the peer dependency warnings at all

packages/create-react-native-library/templates/native-common/android/build.gradle
satya164: hmm this doesn't seem good. i wouldn't feel comfortable having my own library under margelo's domain.

mrousavy: Yea Nitro expects margelo.com namespaces for all Hybrid Objects.
This was to simplify the ProGuard annotations (if something is missing from com.margelo it was stripped out, which we explicitly deal with) - and also it makes cross-referencing Hybrid Objects from other Nitro Modules easier since we don't have to worry about different namespaces here.
Also, it's at least a little bit of branding 😄

mrousavy: > i wouldn't feel comfortable having my own library under margelo's domain.

All the generated code from Turbo Modules are under facebook's domain. This is the same here, the generated code is under margelo domain.

Nitro expects registered hybrid objects to be under the same domain as well, not any custom configurable domain - which is the same as how it works in C++ Turbo Modules.
 This was just to simplify the whole thing.

satya164: > All the generated code from Turbo Modules are under facebook's domain. This is the same here, the generated code is under margelo domain.

Having the generated code or a specific part of the code vs the entire package like it's done here are quite different things. I'm sure bigger companies would have restrictions on using their namespace and publishing to their internal registries.

mrousavy: yea I understand, but this is not gonna change anytime soon as it might require larger changes in the Nitrogen codebase.

mrousavy: If a bigger company reaches out to me because they want to use Nitro but can't because of namespace restrictions, then I'll work on that with them. For now, I don't see a problem in it tbh.

packages/create-react-native-library/templates/native-common/android/build.gradle
atlj: FYI I tried to build nitro without these, but I got an error mentioning some of these libs were duplicated so these are needed for nitro modules right now.

mrousavy:
👀On 07.02.2025, at 01:06, Burak Güner ***@***.***> wrote:﻿
Merged #721 into main.

PR #725, opened 2024-12-08
feat: add support for expo example in native libraries

satya164:


KiwiKilian:
Would this allow two parallel examples? React Native and Expo in parallel? While we use Builder Bob, we ditched the example. Instead we have monorepo setup to have a React Native and Expo example app in parallel, to validate both flavors work. https://github.com/maplibre/maplibre-react-native/tree/807db0e22593cc48fc316dcd9e91c27b1e6186a3/packages

satya164:
@KiwiKilian the CLI will only generate a single example of the chosen type, but you can always copy the example folder to a new folder to have 2 examples.

KiwiKilian:
Only problem with this approach is the code sharing between both examples. But I do understand, that you won't to support this from the CLI creation, as it's rather complex.

packages/create-react-native-library/templates/native-common-example/example/react-native.config.js
```diff
@@ -15,7 +15,7 @@ module.exports = {
       automaticPodsInstallation: true,
     },
   }),
-<% } else { -%>
+<% } else if (example === 'vanila') { -%>
```
KiwiKilian, line 18:
```suggestion
<% } else if (example === 'vanilla') { -%>
```

PR #732, opened 2024-12-19
feat(bob): support custom target definitions

atlj:
### Summary

This adds the `custom` target to bob. Users are able to pass arbitrary scripts via this target and bob will call those scripts.

### Test plan

1. Configure bob in a new project
2. Define the `custom` target
3. Define a script that generates some files
4. Call bob build and make sure the script was called with the right package manager


packages/react-native-builder-bob/src/index.ts
```diff
@@ -584,6 +585,15 @@ async function buildTarget(
         report,
       });
       break;
+    case 'custom':
+      await customTarget({
+        options: targetOptions,
+        source: path.resolve(root, source),
+        output: path.resolve(root, output, 'typescript'),
```
satya164, line 592: TypeScript? 🤔

atlj: They said copying and pasting always works 🙈

packages/react-native-builder-bob/src/targets/custom.ts
```diff
@@ -0,0 +1,57 @@
...
+
+type Options = Omit<Input, 'output'> & {
+  options?: {
+    script?: string;
+    clean?: string;
+  };
+};
+
+export default async function customTarget({ options, root, report }: Options) {
+  if (options?.script === undefined) {
```
satya164, line 16:
```suggestion
  if (options?.script == null) {
```

packages/react-native-builder-bob/src/targets/custom.ts
```diff
@@ -0,0 +1,57 @@
...
+        `No script was provided with the custom target.
+         Example: ${kleur.green('{["custom", { "script": "generateTypes" }}')}`
+      )
+    );
+    process.exit(1);
+  }
+
+  if (options.clean) {
+    report.info(
+      `Cleaning up previous custom build at ${kleur.blue(
```
satya164, line 28:
```suggestion
      `Cleaning up ${kleur.blue(
```

packages/react-native-builder-bob/src/targets/custom.ts
```diff
@@ -0,0 +1,57 @@
...
+  // usr/bin/yarn -> yarn
+  const packageManagerName = path.basename(packageManager);
+  report.info(
+    `Calling ${kleur.blue(packageManagerName)} ${kleur.blue(
+      packageManagerArgs.join(' ')
+    )}`
+  );
+
+  try {
+    await spawn(packageManager, packageManagerArgs, {
```
satya164, line 48: I'd use `npm_execpath` directly instead of only package manager. it means if user runs bob with a specific version of `npm`/`yarn` - we use the same version in case a different version is available globally.

packages/react-native-builder-bob/src/targets/custom.ts
```diff
@@ -0,0 +1,57 @@
...
+      stdio: ['ignore', 'ignore', 'inherit'],
+    });
+  } catch (e) {
+    report.error(`Couldn't run the ${kleur.blue(options.script)} script`);
```
satya164, line 52: Technically we could run it, but it failed with an error
```suggestion
    report.error(`An error occurred when running ${kleur.blue(options.script)}`);
```

packages/react-native-builder-bob/src/targets/custom.ts
```diff
@@ -0,0 +1,57 @@
...
+    await del([path.resolve(root, options.clean)]);
+  }
+
+  const packageManager = process.env.npm_execpath ?? 'npm';
+  const packageManagerArgs = ['run', options.script];
+
+  // usr/bin/yarn -> yarn
+  const packageManagerName = path.basename(packageManager);
+  report.info(
+    `Calling ${kleur.blue(packageManagerName)} ${kleur.blue(
```
satya164, line 42: Since we use the "run" terminology later

```suggestion
    `Running ${kleur.blue(packageManagerName)} ${kleur.blue(
```

packages/react-native-builder-bob/src/targets/custom.ts
```diff
@@ -0,0 +1,57 @@
...
+  try {
+    await spawn(packageManager, packageManagerArgs, {
+      stdio: ['ignore', 'ignore', 'inherit'],
+    });
+  } catch (e) {
+    report.error(`Couldn't run the ${kleur.blue(options.script)} script`);
+    process.exit(1);
+  }
+
+  report.success(`Ran the ${kleur.blue(options.script)} script succesfully`);
```
satya164, line 56: Maybe we should also validate that the folder specified in `clean` option was actually created after the build, in case there's a typo in the name

packages/react-native-builder-bob/src/targets/custom.ts
atlj: This is what's happening. I think the naming was a bit confusing. I've updated it.

PR #736, opened 2024-12-24
fix: bob codegen deletes view specs when codegen type is all

### LINKED ISSUE #728: Building Fabric View + Turbo Module
### Discussed in https://github.com/callstack/react-native-builder-bob/discussions/726

<div type='discussions-op-text'>

<sup>Originally posted by **ian-wd** December  9, 2024</sup>
Is it possible atm to build a Fabric view along with a turbo module?

I was trying to get it working like how [rn-skia](https://github.com/Shopify/react-native-skia/blob/main/packages/skia/src/specs/NativeSkiaModule.ts) does it but I can't get it working via builder-bob

here's a video showing my problem:
https://github.com/user-attachments/assets/499e92ed-7fab-4c37-997b-d0247135b7dc

seems like whenever I add a turbo module spec it doesn't generate the delegate and interface java files</div>

atlj:
### Summary

Fixes #728

The codegen patcher script was deleting the whole `facebook` namespace when the codegen type was all or module. This was not a problem when the type was module but when the type was all, view manager specs were also getting removed.s

### Test plan

1. Create a new turbo module
2. Set codegen type to all
3. Add turbo module specs
4. Run `yarn bob build --target codegen`
5. Make sure `android/generated/java/com/facebook/react` exists and has view manager specs

### DIFF
--- a/packages/react-native-builder-bob/src/targets/codegen/patches/patchCodegenAndroidPackage.test.ts
+++ b/packages/react-native-builder-bob/src/targets/codegen/patches/patchCodegenAndroidPackage.test.ts
@@ -23,12 +23,12 @@ const mockReport: Report = {
   success: console.log,
 };
 
-const mockJavaSpec = `
+const mockJavaModuleSpec = `
 /**
  * Some comment
  */
 
-package com.bobtest;
+package com.facebook.fbreact.specs;
 
 import com.example.exampleimport;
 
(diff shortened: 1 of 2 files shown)


packages/react-native-builder-bob/src/utils/patchCodegenAndroidPackage.ts
```diff
@@ -89,7 +89,19 @@ export async function patchCodegenAndroidPackage(
     })
   );
 
-  await fs.rm(path.resolve(codegenAndroidPath, 'java/com/facebook'), {
-    recursive: true,
-  });
+  if (
+    await fs.pathExists(
+      path.resolve(codegenAndroidPath, 'java/com/facebook/react/viewmanagers')
+    )
```
satya164, line 95: Should we check the `type` instead? What happens you change the `type` from `all`/`view` to module? Does that leave generated view files in the file system?

atlj: @satya164 I just tested it and when the type is `modules`, codegen doesn't generate view specs at all.

satya164: @atlj yeah but my question was about what happens after you change from one type to another. i checked the code and it should be fine since we delete everything https://github.com/callstack/react-native-builder-bob/blob/87b0032acd66b7b14a5b358117a3d1497c59baad/packages/react-native-builder-bob/src/targets/codegen.ts#L32

sidorchukandrew:
Thank you for fixing this!!

sidorchukandrew:
Any updates on when this one can get merged?

atlj:
@sidorchukandrew sorry for the late reply, I'm working on it now 👍

PR #744, opened 2024-12-30
fix: `ndkVersion` casing in `gradle.properties`

KiwiKilian:
### Summary

I think this was a typo? Otherwise it wouldn't match with [this line](https://github.com/callstack/react-native-builder-bob/blob/87b0032acd66b7b14a5b358117a3d1497c59baad/packages/create-react-native-library/templates/native-common/android/build.gradle#L62)?

### Test plan

–


KiwiKilian:
@atlj I've resolved the conflict.

atlj:
Thanks for this

PR #748, opened 2025-01-09
feat: update gh actions in template from `v3` to `v4`

krozniata:
### Summary

Updates GH actions in template from `v3` to `v4`
`v3` use `node v16` under the hood which is deprecated

It also changes `.nvmrc` to use `LTS`

### Test plan

Green CI


atlj (approved):

packages/create-react-native-library/templates/common/$.nvmrc
```diff
@@ -1 +1 @@
-v18
+lts/*
```
atlj, line 1:
```suggestion
v20
```

let's pin this to v20

.nvmrc
```diff
@@ -1 +1 @@
-v18
+lts/*
```
atlj, line 1:
```suggestion
v20
```

PR #753, opened 2025-01-15
fix: build architectures needed only for cpp projects

okwasniewski:
### Summary

This PR adds an additional if statement for defining `reactNativeArchitectures` as its needed only for C++ projects.

### Test plan

CI Green


atlj (approved):
LGTM

PR #759, opened 2025-01-30
chore: adding .xcode.env.local to $.gitignore

Mhp23:
### Summary

Adding `.xcode.env.local` to the `.gitignore`


atlj (approved):
Thanks for this

PR #764, opened 2025-02-07
Generated Prefix is not required

pawanpyakurel:
### Summary

We don't need the generated prefix. 
This is causing the issues.


dylancom:
Does this also remove generated from the podspec? s.private_header_files = "ios/generated/**/*.h"
This was causing issues for me when I started adding Swift code...

Issue #767, opened 2025-02-14
Fabric view giving Cmake error with new arch enabled in React Native 0.77


ShivamJoker:
### Description

Hi I created fabric library with latest version (0.77.1) it's giving me CMake errors.
Have already tried rebuilding and deleting cache 5 times.

<details>
<summary>Expand error log</summary>
BUILD FAILED in 17s
error Failed to install the app. Command failed with exit code 1: ./gradlew app:installDebug -PreactNativeDevServerPort=8081 FAILURE: Build failed with an exception. * What went wrong:
Execution failed for task ':app:configureCMakeDebug[arm64-v8a]'.
[...]

PR #769, opened 2025-02-18
fix: bring back #715 to fix Android builds for Fabric components

Serchinastico:
### Summary

I'm bringing back the fix implemented in https://github.com/callstack/react-native-builder-bob/pull/715. Unfortunately, it was overwritten in https://github.com/callstack/react-native-builder-bob/pull/721 so versions 0.48.x of create-react-native-library are failing Android builds for Fabric components.

Fixes https://github.com/callstack/react-native-builder-bob/issues/767

### DIFF
--- a/packages/create-react-native-library/CHANGELOG.md
+++ b/packages/create-react-native-library/CHANGELOG.md
@@ -3,6 +3,12 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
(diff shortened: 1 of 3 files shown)


atlj:
Thank you so much for the PR and sorry for the delay. I'm checking whether there is a fix without using `type: all` since the RN team asked us to not pass `all` for views since that's reserved for libraries that has views + modules

atlj (approved):
Okay I double checked the RN discord channel and [according to this discussion](https://discord.com/channels/514829729862516747/615932093498654731/1313872017845522462), we can get views back to `all`.
image

Issue #768, opened 2025-02-15
Scaffold a React Native library: Nitro module


mensonones:
### Description

After generated project, i followed steps:

**Get started with the project:**

```bash
yarn
```

**Run the example app on iOS:**

```bash
yarn example ios
```

**Run the example app on Android:**

```bash
yarn example android
```

**I ran the command**:  

```bash 
yarn example android
 ```

I got the error: 

```bash
FAILURE: Build completed with 2 failures.

1: Task failed with an exception.
-----------
* Where:
Build file '/home/jack/PersonalProjects/awesome-library-nitro/android/build.gradle' line: 29

* What went wrong:
A problem occurred evaluating project ':react-native-awesome-library-nitro'.
> Could not read script '/home/jack/PersonalProjects/awesome-library-nitro/nitrogen/generated/android/awesomelibrarynitro+autolinking.gradle' as it does not exist.
```

the folder **nitrogen** was not generated

but since I know a little about Nitro, I already knew what I should do and I accessed it: https://nitro.margelo.com/docs/nitrogen#configuration

after running the command in the root of the project

```bash
yarn nitro-codegen
```
the folder **nitrogen** was generated

and I re-ran the command

```bash 
yarn example android
 ```
I got the error

```bash
BUILD FAILED in 2m 2s
error Failed to install the app. Command failed with exit code 1: ./gradlew app:installDebug -PreactNativeDevServerPort=8081
/home/jack/PersonalProjects/awesome-library-nitro/example/node_modules/react-native-nitro-modules/android/src/main/java/com/margelo/nitro/core/HybridObjectRegistry.java:26: warning: [removal] HybridObjectInitializer in com.margelo.nitro.core has been deprecated and marked for removal public static native void registerHybridObjectConstructor(String hybridObjectName, HybridObjectInitializer initializer); ^
1 warning
e: file:///home/jack/PersonalProjects/awesome-library-nitro/android/src/main/java/com/margelo/nitro/awesomelibrarynitro/AwesomeLibraryNitro.kt:3:2 Unresolved reference 'DoNotStrip'. FAILURE: Build failed with an exception. * What went wrong:
Execution failed for task ':react-native-awesome-library-nitro:compileDebugKotlin'.
> A failure occurred while executing org.jetbrains.kotlin.compilerRunner.GradleCompilerRunnerWithWorkers$GradleKotlinCompilerWorkAction > Compilation error. See log for more details * Try:
[...]

PR #770, opened 2025-02-18
fix: Import `@DoNotStrip`

mrousavy:
### Summary

- Fixes https://github.com/callstack/react-native-builder-bob/issues/768


atlj (approved):
thanks for this!

PR #779, opened 2025-02-26
chore: drop 0.71 support from podspec

okwasniewski:
### Summary

Since React Native 0.75 is dropping support soon, I decided to remove this check for 0.71 this makes the `podspec` a lot cleaner.

### Test plan

CI Green

### DIFF
--- a/packages/create-react-native-library/templates/native-common/{%- project.name %}.podspec
+++ b/packages/create-react-native-library/templates/native-common/{%- project.name %}.podspec
@@ -29,5 +29,11 @@ Pod::Spec.new do |s|
   add_nitrogen_files(s)
 <% } -%>
 
+# Use install_modules_dependencies helper to install the dependencies if React Native version >=0.71.0.
+# See https://github.com/facebook/react-native/blob/febf6b7f33fdb4904669f99d795eba4c0f95d7bf/scripts/cocoapods/new_architecture.rb#L79.
+if respond_to?(:install_modules_dependencies, true)
   install_modules_dependencies(s)
+else
+  s.dependency "React-Core"
+end
 end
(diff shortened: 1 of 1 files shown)


satya164:
Does it affect old architecture? If yes I think we should drop it together with dropping support for old architecture, once new architecture is a bit more stabilized. Normally Bob doesn't drop compatibility unless there's a breaking change that makes it necessary.

Though seems all of that new architecture specific code inside the if condition isn't needed either way so can be dropped.

okwasniewski:
@satya164 No it doesn't `install_modules_dependencies` is used for both old and new arch

satya164:
> @satya164 No it doesn't `install_modules_dependencies` is used for both old and new arch

But isn't that method only available from 0.71? Then the current code does affect both old and new architecture and the change will drop support for old architecture on 0.71 and earlier.

packages/create-react-native-library/templates/native-common/{%- project.name %}.podspec
```diff
@@ -30,29 +29,5 @@ Pod::Spec.new do |s|
...
-      s.dependency "React-RCTFabric"
-<% } -%>
-      s.dependency "React-Codegen"
-      s.dependency "RCT-Folly"
-      s.dependency "RCTRequired"
-      s.dependency "RCTTypeSafety"
-      s.dependency "ReactCommon/turbomodule/core"
-    end
-  end
+  install_modules_dependencies(s)
```
satya164, line 32:
```suggestion
# Use install_modules_dependencies helper to install the dependencies if React Native version >=0.71.0.
  # See https://github.com/facebook/react-native/blob/febf6b7f33fdb4904669f99d795eba4c0f95d7bf/scripts/cocoapods/new_architecture.rb#L79.
  if respond_to?(:install_modules_dependencies, true)
    install_modules_dependencies(s)
  else
    s.dependency "React-Core"
  end
```

okwasniewski: Ah this is what you meant, you want to preserve compatibility for old architecture below 0.71 sounds good to me

satya164 (approved):

PR #806, opened 2025-03-29
feat: validate configuration with arktype

satya164:


atlj (approved):
Tested and works well

PR #810, opened 2025-04-04
fix(crnl): certain versions of XCode can't resolve codegen headers

atlj:
### Summary

- Some people are having XCode build issues with new-architecture enabled libraries, fabric and turbo modules.
- The issues are related to XCode not being able to resolve `#import` directives following this pattern `generated/x/y.h`
- This PR changes the template so these headers are imported from the generated LLVM module using the module include paths such as: `<ModuleName/header.h>`
- Also made all the headers private. Otherwise the modules fail when user tries to add a Swift file. This is because cocoapods automatically generates an umbrella header based on all the available headers and the swift compilation step treats all headers in the umbrella header as objective-c headers. This in turn results in an error since some headers need objective-cpp features and cpp stdlib headers.

### Test plan

1. Create a turbo module and a fabric view separately
2. Make sure you're able to build the library on XCode 16.0, 16.1, and 16.2


satya164 (approved):
Great work!

PR #819, opened 2025-04-11
fix: don't ship generated code with the library

satya164:
### Summary

We currently have `includesGenerateCode: true` which ships codegen specs with the library. The original idea was that the generated code is consistent regardless of React Native version, which can lead to better stability.

However, in a practical sense, a different React Native version may not support this generated code anyway, so the library may get locked to single React Native version anyway.

> The generated code will use the React Native version defined inside your library. So if your library is shipping with React Native 0.76, the generated code will be based on that version. This could mean that the generated code is not compatible with apps using previous React Native version used by the app (e.g. an App running on React Native 0.75). - [Including Generated Code into Libraries](https://reactnative.dev/docs/the-new-architecture/codegen-cli)

This has caused issues for us - breaking our template almost every React Native release due to API changes in the codegen script, which adds a maintenance burden without any concrete benefits.

So, this change removes this option from the default template. The documentation has been updated as well with instructions on how to enable this.

### Test plan

Generate the Turbo Module and Fabric templates and verify that the app builds and runs on Android & iOS.


satya164:
Merging for now since this has been open for a while. I'll address any feedback in another PR.

PR #822, opened 2025-04-15
fix: Migrate to babel-plugin-syntax-hermes-parser

huntie:
### Summary

Fixes compatibility of `react-native-builder-bob/babel-preset` with React Native versions `>= 0.76.0` — likely broken today, without user config modifications.

- `@babel/preset-flow` is [discontinued](https://github.com/babel/babel/issues/16264). The Flow language has since evolved (e.g. `as` syntax) and newer `react-native` versions can no longer be parsed.
- This is succeeded by [babel-plugin-syntax-hermes-parser](https://github.com/facebook/hermes/tree/main/tools/hermes-parser/js/babel-plugin-syntax-hermes-parser).

Context:
- https://github.com/facebook/react-native/issues/50683#issuecomment-2804970384

### Test plan

> [!Note]
> TO DO (would also appreciate if a maintainer had time to E2E validate this)


satya164:
Thanks for the PR. Seems the jest tests are failing after this change with a syntax error in some flow files 🤔

Looking at the related issue, I'm also a bit confused why they have this issue. We use the custom preset only for library code, and still use the React Native babel preset for both app and library code in `src` - so our babel preset shouldn't affect `react-native` in `node_modules` https://github.com/callstack/react-native-builder-bob/blob/7dc0af28fc6dd13d97699b0b8bcf9c254f014a8d/packages/react-native-builder-bob/babel-config.js#L54

Haven't had a chance to try it locally though.

satya164:
Ah, I guess coz `babel-plugin-syntax-hermes-parser` is just a syntax plugin and not a transform. Is there a flow transform plugin?

satya164:
Looks like React Native includes `flow-strip-types` transform https://github.com/facebook/react-native/blob/38fefb2771d7ff65568fa989f2ca4a5bf44e58ec/packages/react-native-babel-preset/src/configs/main.js#L214

Which we also need to include to preserve the previous behavior.

huntie:
@satya164 Ah yeah, need both. Are you happy to own?

satya164:
@huntie yeah I'll take care of adding it.

satya164 (approved):

PR #832, opened 2025-05-04
fix: upgrade to latest react-native

satya164:


satya164:
merging since nitro modules is also failing on main

PR #859, opened 2025-06-05
feat: Enable 16 KB page size on Android

mrousavy:
Enables support for 16 KB page sizes on Android.

Each Native Module library will need to add this for themselves too in their `build.gradle` file if they want to support 16 KB page sizes, but it's good to have it in the template now I think.

See [Support 16 KB page sizes](https://developer.android.com/guide/practices/page-sizes) for more information.


satya164 (approved):
Thanks

PR #863, opened 2025-06-17
fix(template): remove invalid render property from bug report template

saseungmin:
### Summary
Fix bug report issue template not displaying properly due to invalid `render: js` property.
- Removed invalid `render: js` property from bug report template
- Issue template now displays correctly in GitHub Issues interface

#### before
image

#### after

image

#### reference
- https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-githubs-form-schema#input


satya164 (approved):
Thanks for the PR

PR #864, opened 2025-06-17
fix(template): remove the view manager template for iOS

ArturKalach:
### Summary

Currently, with React Native versions `0.79.x` to `0.80.x`, the definition of `Fabric View` components has been changed. You can find that there is no `ViewManager` in the React Native documentation: https://reactnative.dev/docs/next/fabric-native-components-introduction

Additionaly, `RCTViewManager` could lead to the problem with registration:
image
* It could be that the `componentProvider` inside `codegenConfig` fix the problem

I tested it with different versions of React Native and found that `ViewManager` is only needed for the old architecture and React Native versions below `0.74.x`.

_____


mdjastrzebski (approved):
👏

satya164 (approved):
Thank you for the PR

PR #866, opened 2025-06-22
fix: allow project creation in directories with only .git folder

saseungmin:
#### Problem
Currently, the library validation prevents creating projects in directories that contain any files, including `.git` folders. This creates a frustrating user experience when:

1. Users clone an empty GitHub repository
2. The directory contains only a `.git` folder
3. The validation fails with "Directory already exists" error

This is a common workflow where developers create a GitHub repo first, then clone it locally to set up their React Native library.

image

#### Solution
- Treat directories as "empty" if they contain no files OR only a `.git` folder
- Allow project creation in such directories

This follows the same pattern used by Vite's `create-vite` tool:
- [Vite source code](https://github.com/vitejs/vite/blob/b135918b91e8381c50bd2d076d40e9a65fe68bfe/packages/create-vite/src/index.ts#L589)

### Test plan
1. yarn install
2. cd packages/create-react-native-library
3. yarn prepare
4. run `./bin/create-react-native-library`

Tested scenarios:
- [x] Empty directory - works as before
- [x] Directory with only `.git` - now works ✅
- [x] Directory with other files - still blocked as expected
- [x] Non-existent directory - works as before

### DIFF
--- a/packages/create-react-native-library/src/input.ts
+++ b/packages/create-react-native-library/src/input.ts
@@ -210,8 +210,23 @@ export async function createQuestions({
           return 'Cannot be empty';
         }
 
-        if (fs.pathExistsSync(path.join(process.cwd(), input))) {
-          return 'Directory already exists';
+        const targetPath = path.join(process.cwd(), input);
+
+        if (fs.pathExistsSync(targetPath)) {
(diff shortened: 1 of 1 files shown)


satya164 (approved):
Thank you for the PR

Issue #852, opened 2025-05-24
Android Nitro Module wrong package path created


maitrungduc1410:
### Description

I just created a new library with Nitro modules, after running Nitrogen, I start Android project, and it's throwing:

```
/Users/bytedance/Code/Personal/react-native/react-native-app-nitro/example/android/app/build/generated/autolinking/src/main/java/com/facebook/react/PackageList.java:14: error: cannot find symbol
import com.margelo.nitro.appnitro.AppNitroPackage;
                                 ^
  symbol:   class AppNitroPackage
  location: package com.margelo.nitro.appnitro
/Users/bytedance/Code/Personal/react-native/react-native-app-nitro/example/android/app/build/generated/autolinking/src/main/java/com/facebook/react/PackageList.java:62: error: cannot find symbol
      new AppNitroPackage(),
```

I found that it tries to `import com.margelo.nitro.appnitro.AppNitroPackage;`, but the generated project, the AppNitroPackage is in wrong folder:

![Image](https://github.com/user-attachments/assets/0711c1ea-f9d2-4e91-9367-7b582a1789e7)

I manually move the file to the expected folder and able to run Android

### Packages

- [x] create-react-native-library
- [ ] react-native-builder-bob

### Selected options

Nitro Module

### Link to repro

NA

### Environment

Macbook M1

GNUGradyn:
Same issue here. Moving that file and also manually fixing the namespace in that file to match the other file in that folder fixed the issue

PR #869, opened 2025-07-15
fix: fix `…Package.kt` being in wrong Java `package` for Nitro

mrousavy:
### Summary

Previously, CRNL moved the `...Package.kt` into the wrong Java `package`. The Java `package` from the `build.gradle` is expected.
This PR fixes that.

### Test plan

Build a newly created Nitro library with CRNL. Do we have CI for that? We should definitely add one


Nodonisko:
Fixes #852

PR #871, opened 2025-07-16
fix: Remove folly flags that make Nitro Views crash with SIGABRT on RN 0.80

mrousavy:
### Summary

See https://github.com/facebook/react-native/issues/52366#issuecomment-3077780700


satya164 (approved):
Thank you

PR #889, opened 2025-09-24
chore: update iOS TurboModule template

krozniata:
### Summary

Updates iOS `.mm` TurboModule template file to match https://reactnative.dev/docs/turbo-native-modules-introduction#4-write-your-native-platform-code

### Test plan

Module generates correctly, app consuming it builds as expected


satya164 (approved):
Thanks for the PR

PR #891, opened 2025-10-01
chore: Rename nitro-codegen to nitrogen

riteshshukla04:
### Summary
Nitro-codegen is renamed to nitrogen .
This fixes below issue
image

 

### Test plan

N/A

---

> [!NOTE]
> In the template package.json, switch devDependency from `nitro-codegen` to `nitrogen` and remove the `nitrogen` script.
> 
> - **Templates**:
>   - **`packages/create-react-native-library/templates/common/$package.json`**:
>     - Dev dependency: replace `nitro-codegen` with `nitrogen` when `nitro-modules`/`nitro-view` is selected.
>     - Scripts: remove `"nitrogen": "nitro-codegen"` entry.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7c96b49f9dfb1d6934503edda5400f2bd2ed8b8d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #895, opened 2025-10-15
fix: nitrogen script  missing during release-it

riteshshukla04:
### Summary

I did a mistake here https://github.com/callstack/react-native-builder-bob/pull/891. I shouldn't have removed `"nitrogen":"nitrogen" `. 
During Release `npm run nitrogen` is called i think.
This is breaking release it in case of nitro modules. 

 This also fixes one more thing  
I noticed during nitro views the nitrogen wont be created in files in package.json (verified with `npm pack --dry-run`).

Screenshot 2025-10-15 at 1 15 41 PM

Nitro modules require nitrogen to be present in packaged.

This would fix below issue

image

### Test plan

N/A

### DIFF
--- a/packages/create-react-native-library/templates/common/$package.json
+++ b/packages/create-react-native-library/templates/common/$package.json
@@ -18,7 +18,7 @@
     "android",
     "ios",
     "cpp",
-<% if (project.moduleConfig === "nitro-modules") { -%>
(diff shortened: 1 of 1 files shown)


satya164 (approved):
Thanks for the PR

PR #903, opened 2025-10-27
fix: Migrate from `TurboReactPackage` to `BaseReactPackage`

mrousavy:
TurboReactPackage is deprecated by RN

### Summary

See https://github.com/mrousavy/nitro/pull/994


mrousavy:
i need merge queues too this is cool

PR #905, opened 2025-11-10
chore(gitignore): remove `nitrogen` folder

moskalakamil:
### Summary

As Nitro Modules docs recommend, this folder should be tracked in git to improve DX
(https://nitro.margelo.com/docs/nitrogen#2-generate-native-specs)
image

---

> [!NOTE]
> Removes `nitrogen/` from the template `.gitignore` so the folder is tracked.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 862aa8c51d644994367d9b15e3084e813a074844. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/create-react-native-library/templates/common/$.gitignore
+++ b/packages/create-react-native-library/templates/common/$.gitignore
@@ -81,6 +81,3 @@ lib/
 # React Native Codegen
 ios/generated
 android/generated
-
-# React Native Nitro Modules
-nitrogen/


satya164:
Thanks for the PR. However, the note is not entirely correct

> This way your library will always ship a working package as a whole (including generated interfaces), and the user does not need to do anything else than to install your package.

Committing generated code to git is not needed for this to work. The generated code is always generated when publishing the package with the `prepare` script, which makes sure that the library always ships a working package including generated code, and the user doesn't need to do anything.

Committing generated files adds noise to PR reviews - so the only reason you'd commit them is if that's desirable for you and you want to review what changed in the generated code. Typically, it's not common for people to review changes in generated files, but you can always change the default if that's what you want.

So I'm closing the PR.

Issue #912, opened 2025-11-28
Cannot read property 'useContext' of null


vinils:
### Description

problem happened after changing default project to return an element ("Text") mainly on windows

<img width="575" height="911" alt="Image" src="https://github.com/user-attachments/assets/27188b67-f114-47de-ac60-5c99ac3aab5b" />

<img width="506" height="260" alt="Image" src="https://github.com/user-attachments/assets/7a45b1bc-181f-4548-be1b-b2f8ec643223" />

i tested it in different enviroments and OS. was able to make it work only 1 version of ubuntu

### Packages

- [x] create-react-native-library
- [ ] react-native-builder-bob

### Selected options

<img width="632" height="220" alt="Image" src="https://github.com/user-attachments/assets/8199ed5d-ed05-4934-9a4a-12c812de43f7" />

### Link to repro

notpublished

### Environment

@react-native-community/cli": "latest"
nvm -v  1.2.2
npm -v 11.6.2
yarn -v 1.22.22

vinils:
the problem happens its been deployed with react, and react-native on devDependecies of package.json.
Removing it will fix it

vinils:
after digging deeper the problem happened because of different react and react-native dependecies since they are being used in dev and peer dependencies at the same time

PR #914, opened 2025-12-06
Duplicated react and react-native dependencies

vinils:
Problem creating JS library and adding a simple component since its declaring react and react native into devDependecies as issue #912 (https://github.com/callstack/react-native-builder-bob/issues/912)
image

This PR remove/replace react and react-native from dev dependecies by their types once it should be used on peer dependecies only.

### Summary

as issue https://github.com/callstack/react-native-builder-bob/issues/912
while creat-react-native-library -> JSLibrary, sample project breaks while creating any simple component since they use different types of react and react native reference. one for the library and anathoer for the example.

### Test plan

create a JSLibrary template and add a simple component as suggest below:
image

### DIFF
--- a/packages/create-react-native-library/templates/common/$package.json
+++ b/packages/create-react-native-library/templates/common/$package.json
@@ -87,11 +87,9 @@
   },
   "peerDependencies": {
     "react": "*",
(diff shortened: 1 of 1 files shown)


satya164:
I have released a new version that should fix this on Windows. You can try upgrading `react-native-builder-bob` and `react-native-monorepo-config` packages in your project.

